PR #3, opened 2025-03-19
Method processing rfc

KirillPamPam:


docs/rfc/002-UpstreamModule-MethodProcessing.adoc
```diff
@@ -0,0 +1,174 @@
...
+== Detailed design
+
+There is a config file with methods, their quorums and tag parsers.
+[source,yaml]
+----
+methods-config:
+  protocol-methods:
+    eth:
+      - name: "eth_call"
+        quorum: AnyQuorum
```
Termina1, line 25: Is quorum still relevant?

KirillPamPam: I think we need this special logic and it's related to methods, so it depends on how we want to configure it. Either through this config or somehow else.
Also, we can name it differently - not a "Quorum", but a "Strategy", for example.

PR #8, opened 2025-04-04
Base request flow

KirillPamPam:


cmd/dshaltie/main.go
```diff
@@ -3,26 +3,52 @@ package main
...
+	httpServer := server.NewHttpServer(mainCtx, appCtx)
+
+	done := make(chan struct{}, 1)
+
+	sigs := make(chan os.Signal, 1)
+	signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)
+	go func() {
+		sig := <-sigs
+		log.Info().Msgf("got signal %v", sig)
+		done <- struct{}{}
```
a10zn8, line 43: we can just cancel main context here

internal/protocol/response.go
```diff
@@ -3,144 +3,171 @@ package protocol
...
 
 type HttpUpstreamResponse struct {
-	id     interface{}
-	result []byte
-	error  *UpstreamError
+	id           string
+	result       []byte
+	error        *ResponseError
+	responseType RequestType
+	errorBody    []byte
```
a10zn8, line 19: should we store it separatly or we can store it as result?
So just represent whole response body

pkg/utils/channels.go
```diff
@@ -0,0 +1,25 @@
...
+			for _, r := range f(v) {
+				o <- r
+			}
+		}
+		close(o)
+	}()
+	return o
+}
+
+func Map[T any, V any](a <-chan T, f func(T) V) chan V {
```
a10zn8, line 16: let's use https://github.com/samber/lo?tab=readme-ov-file#map

pkg/utils/channels.go
```diff
@@ -0,0 +1,25 @@
+package utils
+
+func FlatMap[T any, V any](a <-chan T, f func(T) []V) chan V {
```
a10zn8, line 3: let's use https://github.com/samber/lo?tab=readme-ov-file#flatmap

pkg/utils/pair.go
```diff
@@ -0,0 +1,18 @@
+package utils
+
+type Pair[F any, S any] struct {
```
a10zn8, line 3: let's use https://github.com/samber/lo?tab=readme-ov-file#t2---t9

pkg/utils/channels.go
KirillPamPam: Will remove it, cause don't use it

pkg/utils/pair.go
KirillPamPam: Thanks a lot!

internal/protocol/response.go
KirillPamPam: Good point, agree with you, it's unnecessary to distinguish them

pkg/utils/channels.go
a10zn8: 😿

PR #12, opened 2025-05-06
Method config rfc v2

KirillPamPam:


docs/rfc/005-MethodConfig.adoc
```diff
@@ -0,0 +1,176 @@
+= RFC Template
```
KirillPamPam, line 1: Fix it according to the new rules

Termina1 (approved):

PR #50, opened 2025-08-28
eth_getLogs blockref parser

Termina1:


internal/caches/cache_policies.go
```diff
@@ -237,6 +238,13 @@ func (c *CachePolicy) isMethodCacheable(ctx context.Context, chain chains.Chain,
 				}
 			}
 		}
+	case *specs.BlockRangeParam:
```
KirillPamPam, line 241: And also we need to handle the finalization case as in the branch above

internal/caches/cache_policies.go
```diff
@@ -237,6 +238,29 @@ func (c *CachePolicy) isMethodCacheable(ctx context.Context, chain chains.Chain,
...
+			return false // not cache requests with block tags
+		}
+		if c.finalizationType == Finalized {
+			chainFinalizedBlock, ok := chainsSupervisor.GetChainState().Blocks[protocol.FinalizedBlock]
+			if ok {
+				var maxBlock int64
+				if param.From == nil {
+					maxBlock = param.To.Int64()
+				} else if param.To == nil {
+					maxBlock = param.To.Int64()
```
KirillPamPam, line 255: Possible nil pointer error?

KirillPamPam (approved):

PR #143, opened 2026-04-01
Send aggregated stats data to DRPC

filinvadim:


internal/caches/cache.go
```diff
@@ -12,13 +12,19 @@ import (
...
+	ErrCacheNotFound  = errors.New("not found in cache")
+	ErrNotImplemented = errors.New("not implemented")
+)
 
 type CacheConnector interface {
 	Id() string
 	Store(ctx context.Context, key string, object string, ttl time.Duration) error
 	Receive(ctx context.Context, key string) ([]byte, error)
 	Initialize() error
+	OutboxStore(ctx context.Context, key string, value []byte, ttl time.Duration) error
```
KirillPamPam: Let's create a new abstraction for these operations. The current `CacheConnector` is responsible to handle RPC requests/responses. I thing it's unnecessary to overload it with this extra logic. Also there might be multiple different `CacheConnector` but for this logic it doesn't make sense. To store and process stats data there must be only one storage and its processor.

internal/caches/cache_policies.go
```diff
@@ -113,6 +113,18 @@ func (c *CachePolicy) Receive(ctx context.Context, chain chains.Chain, request p
 	return object, len(object) > 0
 }
 
+func (c *CachePolicy) OutboxStore(ctx context.Context, key string, value []byte, ttl time.Duration) error {
```
KirillPamPam: The same as above - these operations are not part of `CachePolicy` applicable to RPC requests/responses

internal/caches/postgres_connector.go
```diff
@@ -98,6 +119,69 @@ func (p *PostgresConnector) Initialize() error {
 	return nil
 }
 
+const (
+	createOutboxTable = `
```
KirillPamPam: The same - the specific impl should be outside of this `CacheConnector`. Maybe in that case it's a good idea to rename these cache connectors from `PostgresConnector` to `PostgresCacheConnector`

internal/integration/drpc/drpc_http.go
```diff
@@ -93,6 +95,27 @@ func (s *SimpleDrpcHttpConnector) LoadOwnerKeys(ownerId, apiToken string) ([]*Dr
...
+	ctx, cancel := context.WithTimeout(context.Background(), s.requestTimeout)
+	defer cancel()
+
+	path := fmt.Sprintf("/v1/stats/%s/upload", ownerId)
+
+	response, closeBodyFunc, err := s.makeRequest(
+		ctx, http.MethodPost, apiToken, s.baseUrl+path, bytes.NewBuffer(stats),
+	)
+	if err != nil {
+		return fmt.Errorf("couldn't load keys: %w", err)
```
KirillPamPam: Seems like it's a wrong message

internal/integration/drpc_integration.go
```diff
@@ -49,6 +159,7 @@ func (d *DrpcIntegrationClient) InitKeys(_ string, cfg config.IntegrationKeyConf
 	keyEvents := make(chan keydata.KeyEvent, 100)
 	go d.pollKeys(drpcKeyCfg.Owner.Id, drpcKeyCfg.Owner.ApiToken, keyEvents)
 
+	d.ownerID, d.apiToken = drpcKeyCfg.Owner.Id, drpcKeyCfg.Owner.ApiToken
```
KirillPamPam: There might be multiple owners, so it should't work in this way. A better way to handle it is to store all the keys in this IntegrationClient and their owner with apiTokens. `ApiKey` is part of StatsKey so it should be used to handle this logic. + basically `ApiKey` must be mandatory if drpc stats intergration is enabled, otherwise it's impossible to determine whose data it is.

internal/stats/base_stats_service.go
```diff
@@ -36,35 +47,41 @@ func newStatsDataHolder() *statsDataHolder {
...
 	}
 	for _, result := range requestResults {
 		for {
+			if b.ctx.Err() != nil {
+				return
+			}
 			// use a lock-free approach to write in the map and get a snapshot of the map in parallel
 			holder := b.statsDataHolder.Load()
+
+			if holder.closed.Load() {
```
KirillPamPam, line 86: What's the purpose of `closed`? I think that the next operator where we compare holders should handle everything. Or I'm missing smth?

internal/stats/base_stats_service.go
```diff
@@ -89,28 +106,201 @@ func (b *BaseStatsService) AddRequestResults(requestResults []protocol.RequestRe
...
+				log.Error().Err(err).Msg("stats: cannot flush data")
+			}
+			if err := b.flushUnprocessed(); err != nil {
+				log.Error().Err(err).Msg("failed to flush unprocessed stats")
+			}
+			log.Debug().Msg("stats flush finished")
 		case <-b.ctx.Done():
-			log.Debug().Msg("stopping stats aggregation, flushing remaining data...")
-			stop = true
+			_ = b.flush()
```
KirillPamPam, line 140: Maybe `flush` and `flushUnprocessed` should be processed in a separate goroutine to not block the main loop?

internal/stats/base_stats_service.go
```diff
@@ -89,28 +106,201 @@ func (b *BaseStatsService) AddRequestResults(requestResults []protocol.RequestRe
...
+			log.Error().Err(err).Msg("stats: cannot remove outbox data")
 		}
-		go b.integrationClient.ProcessStatsData(currentAggregatedData.statsAggregatedData)
 	}
+	b.outboxCursor.Add(int64(len(stats)))
+	return nil
+}
+
+func (b *BaseStatsService) storeUnprocessed(data []byte) error {
+	ctx, cancelF := context.WithTimeout(b.ctx, time.Second)
```
KirillPamPam: Is one sec enough?

internal/stats/base_stats_service.go
```diff
@@ -89,28 +106,201 @@ func (b *BaseStatsService) AddRequestResults(requestResults []protocol.RequestRe
...
+	return b.outbox.OutboxStore(ctx, key, compressed, time.Hour*24)
+}
+
+type statsItem struct {
+	key   string
+	value []byte
+}
+
+func (b *BaseStatsService) listUnprocessed() ([]statsItem, error) {
+	ctx, cancelF := context.WithTimeout(b.ctx, time.Second)
```
KirillPamPam: The same

internal/stats/base_stats_service.go
```diff
@@ -89,28 +106,201 @@ func (b *BaseStatsService) AddRequestResults(requestResults []protocol.RequestRe
...
+	}
+	if err != nil {
+		return err
+	}
+
+	return b.storeUnprocessed(unprocessed)
+}
+
+func (b *BaseStatsService) flushUnprocessed() error {
+	ctx, cancelF := context.WithTimeout(b.ctx, time.Second)
```
KirillPamPam: I think this ctx + cancel must be closer to a place where they are used. Because there is `b.listUnprocessed()` between them.

internal/stats/base_stats_service.go
```diff
@@ -89,28 +106,201 @@ func (b *BaseStatsService) AddRequestResults(requestResults []protocol.RequestRe
...
+	defer cancelF()
+
+	stats, err := b.listUnprocessed()
+	if err != nil {
+		return err
+	}
+	if len(stats) == 0 {
+		b.outboxCursor.Store(0)
```
KirillPamPam, line 173: I didn't get why we need outboxCursor, could you elaborate on this pls

internal/stats/base_stats_service.go
filinvadim: Yeah, there was CAS race

internal/stats/base_stats_service.go
filinvadim: Main loop is blocked by ticker. There's no point to spawn more goroutines - there's no win here.

internal/stats/base_stats_service.go
filinvadim: You tell me :)

internal/stats/base_stats_service.go
filinvadim: We get 0-4 records, then 5-10 and so on.  It helps to not accumulate large blobs of data in memory.
Successful unprocessed data flush removes outbox record.
In case of failure of flushing we always reset cursor so we could pickup a record and start again.

internal/integration/drpc_integration.go
filinvadim: Fixed. Please check

internal/config/stats_config.go
```diff
@@ -8,6 +8,7 @@ import (
 
 type StatsConfig struct {
 	Enabled       bool            `yaml:"enabled"`
+	StorageType   string          `yaml:"storage_type"`
```
KirillPamPam: `storage-type` - just to have everything in one style

internal/integration/drpc_integration.go
```diff
@@ -20,10 +24,85 @@ type DrpcIntegrationClient struct {
...
+	for apiKey, items := range itemsPerKey {
+		g.Go(func() error {
+			batch := &api.StatsBatch{
+				Entries: items,
+			}
+			bt, err := proto.Marshal(batch)
+			if err != nil {
+				return fmt.Errorf("couldn't marshal batch %w: %v", ErrStatsDataCorrupted, err)
+			}
+			return d.connector.UploadStats(bt, apiKey)
```
KirillPamPam: `ApiKey` shouldn't be used as an apiToken. Its purpose is not to provider auth, there are already owners and theirs tokens. During key polling we can save apikeys + owner/apitoken entry and then use them.

internal/integration/drpc/drpc_http.go
```diff
@@ -93,6 +95,26 @@ func (s *SimpleDrpcHttpConnector) LoadOwnerKeys(ownerId, apiToken string) ([]*Dr
 	}
 }
 
+func (s *SimpleDrpcHttpConnector) UploadStats(stats []byte, apiToken string) error {
+	ctx, cancel := context.WithTimeout(context.Background(), s.requestTimeout)
+	defer cancel()
+
+	path := "/v1/stats/upload"
```
KirillPamPam: As I remember before there was another endpoint - "/v1/stats/{ownerId}/upload"?

internal/stats/hook/stats_hook.go
```diff
@@ -1,13 +1,19 @@
...
 	"context"
-
+	"github.com/drpcorg/nodecore/internal/outbox"
 	"github.com/drpcorg/nodecore/internal/protocol"
 )
 
+type HookStatsService interface {
+	Start(_ outbox.Storer)
+	Stop(ctx context.Context) error
+	AddRequestResults(requestResults []protocol.RequestResult)
```
KirillPamPam, line 9: As discussed, let's leave only `AddRequestResults`

internal/stats/base_stats_service.go
```diff
@@ -89,28 +111,209 @@ func (b *BaseStatsService) AddRequestResults(requestResults []protocol.RequestRe
...
+	aggregated := current.statsAggregatedData
+	err := b.integrationClient.ProcessStatsData(aggregated)
+	if errors.Is(err, integration.ErrStatsDataCorrupted) {
+		log.Error().Err(err).Msg("stats: stats data corrupted")
+		return err
+	}
+	if err == nil {
+		return nil
+	}
+	return b.storeUnprocessed(aggregated)
```
KirillPamPam, line 164: Is it possible that during `ProcessStatsData` some parts of data might be processed and others might not? In that case we assume that all data haven't been processed and there might be duplications then?

internal/upstreams/upstream_factory.go
```diff
@@ -24,6 +25,12 @@ import (
 	"github.com/samber/lo"
 )
 
+type UpstreamStatsService interface {
+	Start(_ outbox.Storer)
+	Stop(ctx context.Context) error
+	AddRequestResults(requestResults []protocol.RequestResult)
```
KirillPamPam, line 28: Here as well

internal/upstreams/upstream_supervisor.go
```diff
@@ -3,20 +3,25 @@ package upstreams
...
 	"github.com/drpcorg/nodecore/internal/resilience"
-	"github.com/drpcorg/nodecore/internal/stats"
 	choice "github.com/drpcorg/nodecore/internal/upstreams/fork_choice"
 	"github.com/drpcorg/nodecore/pkg/chains"
 	"github.com/drpcorg/nodecore/pkg/utils"
 	"github.com/failsafe-go/failsafe-go"
 	"github.com/rs/zerolog/log"
 )
 
+type UpstreamSupervisorStatsService interface {
```
KirillPamPam: Let's leave only `UpstreamStatsService` since they are in the same package

internal/stats/base_stats_service.go
filinvadim: Yes, Kafka will resolve it

internal/integration/drpc/drpc_http.go
filinvadim: Yes, please check second PR

KirillPamPam (approved):

PR #147, opened 2026-04-09
Shared ws subs

KirillPamPam:
The main issue here is that solana subscriptions don't return the unique ids. It breaks the current logic so it's fixed in this PR:

- now all the WS operations are accumulated with one subId, and unsub happens inly when there are no operations at all
- fixed different races in WS operations handling - now all the commands are gone through one channel pipeline
- now solana and eth subs return internal subIds to the clients in order to close the correct WS operation during unsub


internal/upstreams/flow/sub_processor.go
```diff
@@ -29,6 +36,18 @@ func (s *SubscriptionRequestProcessor) ProcessRequest(
 		defer close(responses)
 		var response *protocol.ResponseHolderWrapper
 
+		if request.SpecMethod() == nil && request.SpecMethod().Subscription == nil {
```
tonatoz: Should be || or will get a pinic

internal/upstreams/ws/request_registry.go
```diff
@@ -82,65 +100,55 @@ func (b *BaseRequestRegistry) Start(req RequestOperation, doOnCloseFunc DoOnClos
...
-		subs:       utils.NewCMap[string, RequestOperation](),
+		commands:   make(chan registryCommand, 256),
+		registryState: &registryState{
+			requests: make(map[string]RequestOperation),
+			subs:     make(map[string]*registrySubscription),
+		},
+		allOps: utils.NewCMap[string, RequestOperation](),
 	}
+
+	go registry.run()
```
tonatoz, line 152: Never stop

internal/protocol/response.go
```diff
@@ -29,6 +29,76 @@ type SubscriptionResultResponse struct {
 	result []byte
 }
 
+type SubscriptionMethodResulResponse struct {
```
tonatoz: SubscriptionMethodResultResponse missing T

internal/upstreams/ws/request_registry.go
```diff
@@ -82,65 +100,55 @@ func (b *BaseRequestRegistry) Start(req RequestOperation, doOnCloseFunc DoOnClos
...
+
+	return registry
+}
+
+func (b *BaseRequestRegistry) closeReq(req RequestOperation) {
+	done := make(chan bool, 1)
+	b.commands <- newFinishCommand(req, done)
+	if <-done {
```
tonatoz: deadlock in case of commands filled with 256 elements

tonatoz (changes requested):

tonatoz:
`
  run() goroutine                Start() goroutine              External consumer
        |                                                              |                                                        |
        |--Write(MessageInternal)-->             |                                                        |
        |   [internalMessages buf=50]             |--Write(MessageResponse)-->   |
        |                                                              |   [responseChan buf=50]            | reads here`
        The single
     run() loop executes rpcCommand/subscriptionCommand handlers synchronously, and those handlers do
     blocking writes into per-request internalMessages buffers of size 50. Once one slow subscriber fills
     its buffer, the command loop blocks; after that, closeReq() waits forever for a finishCommand, and
     the 256-slot commands channel eventually fills, freezing unsubscribe/disconnect handling too. The
     shared-subscription fanout makes this much easier to trigger because one blocked op now blocks every
     other op sharing that upstream bucket

internal/upstreams/ws/request_registry.go
KirillPamPam: Fixed

internal/upstreams/ws/request_registry.go
KirillPamPam: Fixed

tonatoz (approved):

PR #161, opened 2026-04-27
Update Chart documentation & validation for "image"

themetalfleece:
**Description**

The chart ([deployment.yaml](https://github.com/drpcorg/nodecore/blob/main/chart/nodecore/templates/deployment.yaml#L47), [values.schema.json](https://github.com/drpcorg/nodecore/blob/main/chart/nodecore/values.schema.json#L6)) uses `image.name` instead of `image.tag` & `image.repository` (which are documented in the README). I learnt it the hard way during a rollout restart :smiling_face_with_tear: 

I've also added `"additionalProperties": false` to the nested json schema to avoid such misconfiguration in the future (my `image.tag` config passed validation but was dropped silently), but let me know if this is not something you want. It'd be nice if this was added to other nested schemas too.

### DIFF
--- a/chart/nodecore/README.md
+++ b/chart/nodecore/README.md
@@ -57,8 +57,7 @@ Create your own `values.yaml`:
 
 ```yaml
 image:
-  repository: drpcorg/nodecore
-  tag: "latest"
+  name: drpcorg/nodecore:latest
   pullPolicy: IfNotPresent
 
 replicaCount: 1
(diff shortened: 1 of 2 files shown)


mxssl:
@themetalfleece Hi! Thank you for your contribution, the issue has been fixed in https://github.com/drpcorg/nodecore/pull/211

PR #179, opened 2026-05-07
ETH label detectors

KirillPamPam:


internal/upstreams/chains_specific/evm_chain_specific_test.go
```diff
@@ -24,7 +24,7 @@ func TestChainValidator(t *testing.T) {
 	}
 	connector := mocks.NewConnectorMock()
 
-	validators := specific.NewEvmChainSpecific("id", connector, chains.UnknownChain, options).SettingsValidators()
+	validators := specific.NewEvmChainSpecific(context.Background(), "id", connector, chains.UnknownChain, options).SettingsValidators()
```
filinvadim, line 27: t.Context()
linter would fail here

filinvadim (approved):

internal/upstreams/chains_specific/evm_chain_specific_test.go
KirillPamPam: As far as I see it's ok for linter

PR #193, opened 2026-05-26
feat: load extra chain definitions and method specs at startup

### LINKED ISSUE #186: feat: allow loading external chain definitions at startup (e.g. NODECORE_EXTRA_CHAINS_PATH)
## Use case

We run nodecore in front of EVM-compatible upstreams that aren't in `drpcorg/public` and never will be — they're customer-specific private/consortium networks (e.g. Hyperledger Besu running with customer-assigned chain IDs). The list of public chains in `pkg/chains/public/chains.yaml` is exactly right as a default, but there's currently no way to extend it at runtime.

## Why a config-only workaround isn't enough

We tried:

- `chain: ethereum` + `options.disable-chain-validation: true` — works for most methods, but `eth_chainId` and `net_version` are marked `"settings": {"local": true}` in `pkg/methods/specs/eth.json` and answered locally from the chain registry. The response is the *spoofed* chain's id (e.g. `0x1`), not the upstream's actual id. This breaks EIP-155 transaction signing for any client that does `eth_chainId` discovery (ethers / viem / web3.js / hardhat / foundry), which is essentially every modern client.
- `methods.disable: [eth_chainId]` per-upstream — doesn't override the `local: true` behaviour. The method is still answered from the chain registry.

So with the current code there's no way to serve `eth_chainId` correctly for a chain that isn't in the embedded `chains.yaml`.

## Proposal

Mirror the existing pattern in `pkg/methods` (`NewMethodSpecLoaderWithFs(fs.FS)`) on the chains side:

1. Refactor `pkg/chains/chains.go` so the embedded `chainsCfg` parse step can also merge an optional external YAML blob (same schema as `drpcorg/public/chains.yaml`).
2. In `cmd/nodecore/main.go`, look for `NODECORE_EXTRA_CHAINS_PATH` at startup; if set, read the file and merge.
3. Document the env var and a minimal example in `docs/nodecore/05-upstream-config.md`.

Net effect: a deployer who needs private chains drops a small `extra-chains.yaml` next to the binary, sets the env var, and nodecore behaves correctly (`eth_chainId` returns the configured id, method specs inherit via `method-spec: eth`, etc.) — no fork required.

Optionally also wire up `NODECORE_EXTRA_SPECS_PATH` symmetrically since `NewMethodSpecLoaderWithFs` is already exported but unused — would let private chains override method specs too if needed.

## Happy to send the PR

Estimated ~200 LOC + tests. Wanted to check the design lands before opening it. Specifically:

- Env var vs config-file field — preference?
- Strict merge (error on conflict) vs override (extra wins) vs append-only (no overlapping `short-names` / `chain-id`)?

robbeverhelst:
Closes #186

## What

Two opt-in env vars that let deployers extend nodecore's embedded chain
registry and method specs at startup, without forking the image:

| Env var | Purpose |
|---------|---------|
| `NODECORE_EXTRA_CHAINS_PATH` | Path to a YAML file using the same schema as `pkg/chains/public/chains.yaml`. Merged on top of the embedded registry. |
| `NODECORE_EXTRA_SPECS_PATH` | Directory of JSON method-spec files (same schema as `pkg/methods/specs/*.json`). Wires up the already-exported `NewMethodSpecLoaderWithFs` constructor. |

If both vars are unset, behaviour is identical to today.

## Why

The motivating use case is documented in #186: running nodecore in
front of EVM-compatible upstreams that aren't in `drpcorg/public` and
never will be — customer-private Besu / consortium networks with their
own chain ids.

The existing per-upstream `disable-chain-validation` flag isn't enough
because `eth_chainId` and `net_version` are marked `"settings": {"local": true}`
in `eth.json` and answered locally from the chain registry. With the
chain spoofed as `ethereum`, those return `0x1` instead of the real
chain id, which breaks EIP-155 signing in every modern client
(ethers / viem / web3.js / hardhat / foundry).

With this PR, a deployer drops a small `extra-chains.yaml` containing
their private chain entry, sets `NODECORE_EXTRA_CHAINS_PATH`, and the
registry returns the correct chain-id end-to-end.

### `pkg/chains/chains.go`

- Renamed `configureChains` → `configureChainsFromBytes(raw []byte)` so the
  same code path handles the embedded YAML and any extra YAML.
- New public `LoadExtraChains(yamlBytes []byte) error`. Conflict checks:
  - Empty input is a no-op.
  - Duplicate `short-name` against an already-registered chain → error.
  - Non-zero `grpcId` collision against an already-registered chain → error.
- For each extra short-name not in `chainsMap`, allocate a `Chain` id
  starting at `dynamicChainBaseId` (`1 << 30`) so it doesn't collide
  with the codegen-emitted constants, and register it in
  `dynamicChainNames` so `Chain.String()` round-trips correctly.

### `cmd/chains/init_chains.go` (code-generator template)

- Declares `const dynamicChainBaseId Chain = 1 << 30` and
  `var dynamicChainNames = map[Chain]string{}` in the generated file.
- `Chain.String()` falls through to `dynamicChainNames[c]` for ids in
  the dynamic range.

The generated `pkg/chains/chains_data.go` is gitignored as today; it's
re-generated next time `make chains` runs.

### `cmd/nodecore/main.go`

- Reads `NODECORE_EXTRA_CHAINS_PATH` after `flag.Parse()` and before
  `config.NewAppConfig()`; calls `chains.LoadExtraChains` if set.
- Reads `NODECORE_EXTRA_SPECS_PATH` next and uses
  `specs.NewMethodSpecLoaderWithFs(os.DirFS(path))` instead of the
  embedded loader if set.

### Docs

New subsection in `docs/nodecore/05-upstream-config.md` covering both
env vars, the schema, and a minimal Besu example.

### Tests

`pkg/chains/chains_extra_test.go`:

- happy path (registers a new chain, `Chain.String()` round-trips,
  `eth` method-spec inherited)
- empty input is a no-op
- duplicate short-name rejected
- grpcId collision rejected
- malformed YAML returns an error

All existing tests still pass; `go vet` and `go build ./...` clean.

## Open questions (carried over from #186)

1. **Env var vs config-file field** — I went with env var to mirror how
   `internal/config/config.go` already does `ConfigPathVar`. Happy to
   move it under `app-config` if you prefer.
2. **Merge semantics** — strict: extras can only *add* new chains, not
   override existing ones. Felt like the safe default; I can switch to
   "extras win" or make it configurable if you'd rather.
3. **`net-version` derivation** — extras follow the same fall-through
   as embedded entries (explicit `net-version` else big-int decimal of
   `chain-id`). Worth calling out in case private chains have weird
   net versions.

Happy to iterate on any of the above.

---

Recreated from the SettleMint fork because the original fork backing #187 was deleted, which closed that PR.

### DIFF
--- a/cmd/nodecore/main.go
+++ b/cmd/nodecore/main.go
@@ -10,7 +10,7 @@ import (
 
 	"github.com/drpcorg/nodecore/internal/app"
 	"github.com/drpcorg/nodecore/internal/config"
-	_ "github.com/drpcorg/nodecore/pkg/chains"
+	"github.com/drpcorg/nodecore/pkg/chains"
 	_ "github.com/drpcorg/nodecore/pkg/errors_config"
 	_ "github.com/drpcorg/nodecore/pkg/logger"
 	specs "github.com/drpcorg/nodecore/pkg/methods"
@@ -18,14 +18,42 @@ import (
 	_ "go.uber.org/automaxprocs"
 )
 
+const (
+	// envExtraChainsPath points at an additional chain-registry YAML (same
+	// schema as drpcorg/public chains.yaml) that gets merged into the
+	// embedded registry at startup. Empty/unset = embedded only.
+	envExtraChainsPath = "NODECORE_EXTRA_CHAINS_PATH"
+
+	// envExtraSpecsPath points at a directory of JSON method-spec files that
+	// extend or override the embedded specs. Empty/unset = embedded only.
+	envExtraSpecsPath = "NODECORE_EXTRA_SPECS_PATH"
+)
+
 func main() {
 	flag.Parse()
+
+	if path := os.Getenv(envExtraChainsPath); path != "" {
+		extra, err := os.ReadFile(path)
+		if err != nil {
+			log.Panic().Err(err).Str("path", path).Msg("unable to read extra chains file")
+		}
+		if err := chains.LoadExtraChains(extra); err != nil {
+			log.Panic().Err(err).Str("path", path).Msg("unable to merge extra chains")
+		}
+		log.Info().Str("path", path).Msg("loaded extra chain definitions")
+	}
 
 	appConfig, err := config.NewAppConfig()
 	if err != nil {
 		log.Panic().Err(err).Msg("unable to parse the config file")
 	}
-	err = specs.NewMethodSpecLoader().Load()
+
+	specLoader := specs.NewMethodSpecLoader()
+	if path := os.Getenv(envExtraSpecsPath); path != "" {
+		specLoader = specs.NewMethodSpecLoaderWithFs(os.DirFS(path))
+		log.Info().Str("path", path).Msg("loading method specs from external directory")
+	}
+	err = specLoader.Load()
 	if err != nil {
 		log.Panic().Err(err).Msg("unable to load method specs")
 	}
--- a/cmd/nodecore/main.go
+++ b/cmd/nodecore/main.go
@@ -23,10 +23,6 @@ const (
 	// schema as drpcorg/public chains.yaml) that gets merged into the
 	// embedded registry at startup. Empty/unset = embedded only.
 	envExtraChainsPath = "NODECORE_EXTRA_CHAINS_PATH"
-
-	// envExtraSpecsPath points at a directory of JSON method-spec files that
-	// extend or override the embedded specs. Empty/unset = embedded only.
-	envExtraSpecsPath = "NODECORE_EXTRA_SPECS_PATH"
 )
 
 func main() {
@@ -49,9 +45,9 @@ func main() {
 	}
 
 	specLoader := specs.NewMethodSpecLoader()
-	if path := os.Getenv(envExtraSpecsPath); path != "" {
-		specLoader = specs.NewMethodSpecLoaderWithFs(os.DirFS(path))
-		log.Info().Str("path", path).Msg("loading method specs from external directory")
+	if path := os.Getenv(specs.SpecPathVar); path != "" {
+		specLoader = specs.NewMethodSpecLoaderWithExtraFs(os.DirFS(path))
+		log.Info().Str("path", path).Msg("extending method specs with external directory")
 	}
 	err = specLoader.Load()
 	if err != nil {
--- a/pkg/chains/chains.go
+++ b/pkg/chains/chains.go
@@ -158,9 +158,13 @@ func loadExtraChainsLocked(extraYaml []byte) error {
 	}
 
 	for shortName, configured := range extraConfigured {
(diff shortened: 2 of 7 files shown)


KirillPamPam:
@robbeverhelst Could you pls pull the latest changes from `main` since there have been a few major fixes/improvements, especially related to method specs

pkg/chains/chains.go
```diff
@@ -93,15 +95,83 @@ var UnknownChain = &ConfiguredChain{
...
+}
+
+func loadExtraChainsLocked(extraYaml []byte) error {
+	extraConfigured, extraGrpc, err := configureChainsFromBytes(extraYaml)
+	if err != nil {
+		return err
+	}
+
+	for shortName, configured := range extraConfigured {
+		if existing, exists := chains[shortName]; exists && existing != configured {
```
KirillPamPam: Is `existing != configured` necessary? As far as I understand if just the same name exists, then throw an error

pkg/chains/chains.go
```diff
@@ -93,15 +95,83 @@ var UnknownChain = &ConfiguredChain{
...
+			return fmt.Errorf("extra chain short-name %q already registered", shortName)
+		}
+		if _, exists := grpcChains[configured.GrpcId]; exists && configured.GrpcId != 0 {
```
KirillPamPam, line 168: Also about `configured.GrpcId != 0`?

cmd/nodecore/main.go
```diff
@@ -10,22 +10,50 @@ import (
...
+
 	appConfig, err := config.NewAppConfig()
 	if err != nil {
 		log.Panic().Err(err).Msg("unable to parse the config file")
 	}
-	err = specs.NewMethodSpecLoader().Load()
+
+	specLoader := specs.NewMethodSpecLoader()
+	if path := os.Getenv(envExtraSpecsPath); path != "" {
+		specLoader = specs.NewMethodSpecLoaderWithFs(os.DirFS(path))
```
KirillPamPam: What about not override but extend the existing specs as in the same way with chains?

pkg/chains/chains.go
robbeverhelst: Removed, redundant.

pkg/chains/chains.go
robbeverhelst: Kept — `grpcId` is optional and 0 means unset (all embedded chains use a non-zero id), so only non-zero ids can collide. Can require an explicit id instead if you prefer.

cmd/nodecore/main.go
robbeverhelst: Done — extras now load on top of the embedded specs (add-only, dup name errors), wired to the existing `NODECORE_SPECS_PATH` const.

robbeverhelst:
Rebased on latest main, addressed the three inline comments.

PR #196, opened 2026-05-28
REST support

KirillPamPam:
# REST routing across HTTP and gRPC

## Summary

Brings REST support up to parity with JSON-RPC across both transports:

- **HTTP server** now extracts the canonical method template, path params, headers, and query from the incoming request through a single parser, exactly the way the gRPC adapter already received them pre-structured.
- **HTTP connector** drops the `if-REST / if-JSON-RPC` type-asserts in the hot path — `SendRequest` dispatches into two focused methods (`sendJsonRpc` / `sendRest`) that share only a small `dispatch` helper for execute-and-receive.
- **Upstream response headers** now flow back to the client on every reply shape (HTTP unary, HTTP stream, gRPC unary, gRPC stream, gRPC error). Previously they were dropped on the gRPC unary-success path.
- **REST non-2xx parsing** gains a three-tier fallback (structured JSON → HTTP status → truncated text) so a multi-kilobyte 502 HTML page no longer poisons logs or surfaces a meaningless 500.
- **Test coverage** added across `pkg/utils`, `pkg/methods`, `internal/protocol`, `internal/server/http_server`, `internal/server/emerald`, `internal/upstreams/connectors`, and `internal/config`.

## Motivation

REST chains (Algorand, Aztec, Hyperliquid, …) need the same first-class routing machinery EVM JSON-RPC has had: spec-driven method identification for stats / caching / method-ban, header + query pass-through, response-header propagation for things like quorum signature verification, and a predictable error contract on non-2xx responses. Pre-change, the HTTP path was forwarding the literal URL as the "method" name, the gRPC adapter was baking query into the path, and 5xx HTML pages were collapsing to generic `500: internal server error`.

### Routing primitives

- **`pkg/utils/path.go`** — new `PathMatcher` (trie, immutable after construction, no mutex, no dedup set) and inverse `BuildRestURL(template, pathParams)` that reconstructs the literal verb + path at send time. Templates look like `GET#/v2/accounts/*` with `*` as the single-segment wildcard; the HTTP verb is hoisted to the first trie segment so `GET` and `POST` on the same path live on separate branches automatically. Literal children beat wildcard siblings at the same level (no backtracking).
- **`pkg/methods/helpers.go` + `resolved_specs.go`** — built a per-spec matcher during `enrichSpec` from every method name containing `#`. Exposed via `specs.MatchRestMethod(specName, fullPath)`.

### Protocol layer

- **`internal/protocol/data.go`** — `RequestParams.Headers` and `QueryParams` switched to `map[string][]string` so repeated keys (`?token=A&token=B`, multi-valued headers) survive end-to-end.
- **`internal/protocol/rest_request.go`** — `UpstreamRestRequest` now carries only the method template; verb + literal path are reconstructed by the connector from the template + `PathParams`. Internal constructors auto-split `?query` from any path they're handed into `RequestParams.QueryParams`, so `Method()` is always a clean template — even for synthetic probes (`/v2/blocks/42?header-only=true`).
- **`internal/protocol/parse_response.go`** (split from `response.go`) — new `parseRestErrorBody` with three tiers:
  1. Body matches `{"code":N,"message":"..."}` → use both; missing `code` falls back to HTTP status.
  2. Body is anything else (text, HTML, malformed JSON, JSON without error fields) → HTTP status as code, body truncated to 512 chars with `... (truncated)` marker.
  3. Empty / whitespace-only → deterministic placeholder.
- The shared `parseError` was renamed to `parseJsonRpcError` (REST no longer routes through it).
- Treats any 2xx as success (`200`/`201`/`204`/…), where the previous `!= 200` check mis-classified non-200 successes as errors.

### HTTP server

- **`internal/server/http_server/rest_parser.go`** — new `parseRestRequest(req, restPath, specName)` builds `VERB#/path`, calls `specs.MatchRestMethod`, falls back to the literal template when the spec has no REST routes, copies headers verbatim, and filters reserved query keys.
- **`internal/server/http_server/reserved.go`** — single source of truth for query keys nodecore intercepts (`quorum`, `quorum_required` today).
- **`internal/server/http_server/handlers.go`** — `NewRestHandler(preReq, *http.Request, restPath)` reads body, validates JSON when non-empty, and feeds the structured params into `NewUpstreamRestRequest(id, methodTemplate, requestParams, body, specName)`.
- **`internal/server/http_server/http_server.go`** — response-header copy extracted into `copyUpstreamResponseHeaders(dst, response)` keyed on the `protocol.HasResponseHeaders` interface (was a concrete `*BaseUpstreamResponse` type assertion).

### gRPC server

- **`internal/server/emerald/native_call_adapter.go`** — `restNativeCallAdapter.BuildRequest` plumbs `restData.PathParams / Headers / QueryParams` straight into `RequestParams`. Validation replaces the old verb/path split (`validateRestMethodTemplate`).
- **`internal/server/emerald/grpc_blockchain.go`** — `nativeCallSuccessItems` now takes `http.Header` and emits `ResponseHeaders` on both single-item and multi-chunk success replies, closing the gap where unary REST success silently dropped headers (streaming + error paths already had them).

### HTTP connector

- **`internal/upstreams/connectors/http_connector.go`** — `SendRequest` is a 2-line dispatch into `sendJsonRpc` (POST to endpoint, body verbatim) or `sendRest` (`BuildRestURL` → join with endpoint → layer `RequestParams.QueryParams` → layer quorum → apply config + client headers). A shared `dispatch` helper handles `httpClient.Do` + stream-vs-buffer framing, parameterized by a `canStreamFunc` (`alwaysStream` for REST, `jsonRpcCanStream` for JSON-RPC). Connector-config headers (auth tokens) win over per-request client headers on conflict.


internal/upstreams/connectors/http_connector.go
```diff
@@ -111,50 +111,145 @@ func (h *HttpConnector) SubscribeStates(_ string) *utils.Subscription[protocol.S
...
+		req.Header.Set(k, v)
 	}
+}
+
+// applyClientHeaders forwards per-request client headers onto the upstream
+// request, except keys the connector config already owns - those are
+// typically auth tokens that a curious client must not be able to override.
+func (h *HttpConnector) applyClientHeaders(req *http.Request, headers map[string][]string) {
+	for k, vs := range headers {
+		if _, taken := h.additionalHeaders[k]; taken {
```
tonatoz: **Header case-sensitivity bug — security-relevant.**

`h.additionalHeaders` is a `map[string]string` populated from YAML; keys preserve whatever case the operator wrote. `headers` originate from `cloneHeaders(req.Header)` and arrive with canonical casing applied by `net/http`. If the config uses `authorization:` (lowercase) and the client sends `Authorization`, this lookup misses, the client header gets added via `req.Header.Add`, and the resulting request carries both the config-supplied token and the client-supplied one.

`TestRestRequest_ConfigHeadersWinOverClientHeaders` only exercises canonical keys and does not catch this.

Canonicalize `additionalHeaders` once in the constructor (or compare via `http.CanonicalHeaderKey` on both sides). Add a regression test using a non-canonical config key.

internal/server/http_server/http_server.go
```diff
@@ -281,6 +233,23 @@ func handleResponse(
 	return writeResponse(httpResponse, code, responseReader)
 }
 
+// copyUpstreamResponseHeaders forwards an upstream's response headers onto
+// the outgoing HTTP response. Used so REST clients see the upstream's
+// Content-Type, quorum signature headers, CORS hints, etc. verbatim.
+// Responses that don't implement HasResponseHeaders (e.g. ReplyError)
+// silently contribute nothing.
+func copyUpstreamResponseHeaders(dst http.Header, response protocol.ResponseHolder) {
```
tonatoz, line 241: **Forwarding every upstream header verbatim is unsafe.**

Hop-by-hop headers (`Transfer-Encoding`, `Connection`, `Keep-Alive`), `Content-Length`, `Content-Encoding`, and `Set-Cookie` get copied to the client response unconditionally. This breaks downstream middleware that rewrites the body (compression layer seeing a stale `Content-Encoding`, length mismatch when framing changes) and leaks upstream cookies to clients with no opt-in.

Introduce an explicit deny-list at minimum: `Transfer-Encoding`, `Connection`, `Keep-Alive`, `Content-Length`, `Content-Encoding`. `Set-Cookie` needs a deliberate policy call — do not default to forwarding it.

pkg/methods/specs/hyperliquid.json
```diff
@@ -0,0 +1,15 @@
+{
```
tonatoz, line 1: **Filename typo.**

Spec content registers the name `hyperliquid`; this file is `hypeliquid.json` (missing `r`). The sibling file `hyperliquid-rest-additional.json` is spelled correctly. Rename to `hyperliquid.json` before this merges — fixing typos in spec filenames after the fact is noisier than fixing them now.

internal/protocol/json_rpc_request.go
```diff
@@ -76,10 +76,8 @@ func NewInternalSubUpstreamJsonRpcRequest(method string, params any, chain chain
 	}, nil
 }
 
-func NewUpstreamJsonRpcRequest(id string, jsonRpcRequest JsonRpcRequestBody, isSub bool, specMethod *specs.Method) *UpstreamJsonRpcRequest {
-	if specMethod == nil {
-		specMethod = specs.DefaultMethod(jsonRpcRequest.Method)
-	}
+func NewUpstreamJsonRpcRequest(id string, jsonRpcRequest JsonRpcRequestBody, isSub bool, specName string) *UpstreamJsonRpcRequest {
+	specMethod := specs.GetSpecMethodWithFallback(specName, jsonRpcRequest.Method)
```
tonatoz, line 80: **Per-request spec lookup on the JSON-RPC hot path.**

Previously the resolved `*specs.Method` was passed in by the caller, so a batch of N items resolved the spec once at the chain-supervisor level. With the new shape a 100-item batch performs 100 map lookups against the resolved-spec registry.

Confirm the resolved-spec map is read under `RLock` (or lock-free) and that the lookup is not on a path that holds any contended lock. If it is, hoist the lookup back to the caller or memoize per `(specName, method)` for the duration of a request batch.

internal/upstreams/connectors/http_connector.go
```diff
@@ -168,21 +263,17 @@ func (h *HttpConnector) SendRequest(ctx context.Context, request protocol.Reques
 
 	if request.IsStream() && resp.StatusCode == 200 && !quorumRequested {
```
tonatoz: **Streaming refuses any 2xx status other than 200.**

This PR explicitly fixes REST success classification to treat the full 2xx range as success (`TestRestRequest_AllTwoHundredCodesAreSuccess`), but this branch still pins on `== 200`. A REST upstream returning 201/206 with a streamable body silently falls through to `receiveWholeResponse`, defeating the streaming intent.

Loosen the condition to `resp.StatusCode >= 200 && resp.StatusCode < 300`. Add a test mirroring the 2xx-success matrix for the stream path.

tonatoz (changes requested):
Five issues to address before this can merge. Inline comments below.

internal/upstreams/connectors/http_connector.go
KirillPamPam: True, thanks!
Fixed

internal/server/http_server/http_server.go
KirillPamPam: Agree, fixed

internal/protocol/json_rpc_request.go
KirillPamPam: It's ok, the algo is lock free

internal/upstreams/connectors/http_connector.go
KirillPamPam: Agree, fixed

tonatoz (approved):

PR #204, opened 2026-06-04
grpc selector routing

Termina1:


internal/config/option_defaults.go
```diff
@@ -119,6 +119,13 @@ func setOptionsDefaults(
 			lo.Ternary(upstreamMode == StrictMode, false, true),
 		)
 	}
+	if upstreamOptions.DisableSafeBlockDetection == nil {
+		defaultValue := getBool(defaultChainOptions, func(options *chains.Options) *bool { return options.DisableSafeBlockDetection })
```
KirillPamPam:
```suggestion
		upstreamOptions.DisableSafeBlockDetection = resolveBool(
			getBool(defaultChainOptions, func(options *chains.Options) *bool { return options.DisableSafeBlockDetection }),
			getBool(globalChainOptions, func(options *chains.Options) *bool { return options.DisableSafeBlockDetection }),
			lo.Ternary(upstreamMode == StrictMode, false, true),
		)
```

KirillPamPam: Let's simplify this part as it's done with other bools

internal/protocol/data.go
```diff
@@ -184,6 +187,125 @@ type SubscriptionResponseHolder interface {
...
+	Reason string
+}
+
+func (RequestUnsupportedSelector) isRequestSelector() {}
+
+type SelectorCarrier interface {
+	Selectors() []RequestSelector
+}
+
+type RequestHolderWithSelectors struct {
```
KirillPamPam: Let's add selectors to the current `RequestHolder`:

- it will simplify the model
- I think that any request (json-rpc, rest, etc) should have selectors

KirillPamPam: \+ no need to wrap/unwrap

internal/upstreams/blocks/block_processor.go
```diff
@@ -96,6 +96,9 @@ func (b *EthLikeBlockProcessor) Start() {
 	b.lifecycle.Start(func(ctx context.Context) error {
 		go func() {
 			b.poll(protocol.FinalizedBlock)
+			if b.safeBlockDetectionEnabled() {
```
KirillPamPam: I believe that each detection should be done separately so that they do not influence each other

internal/upstreams/flow/execution_flow.go
```diff
@@ -159,9 +159,16 @@ func (e *BaseExecutionFlow) AddHooks(hooks ...any) {
 
 func (e *BaseExecutionFlow) createStrategy(ctx context.Context, request protocol.RequestHolder) UpstreamStrategy {
 	chainSupervisor := e.upstreamSupervisor.GetChainSupervisor(e.chain)
+	additionalMatchers := make([]Matcher, 0)
+	var order UpstreamOrder
+	if carrier, ok := request.(protocol.SelectorCarrier); ok {
```
KirillPamPam: I think it's better just to have Selectors() in the RequestHolder interface as mentioned above

internal/upstreams/flow/match_trace.go
```diff
@@ -0,0 +1,39 @@
+package flow
+
+import (
+	"fmt"
+	"strings"
+)
+
+const maxMatchTraceCauseLen = 200
+
+type UpstreamMatchTrace struct {
```
KirillPamPam, line 10: Let's add some tests

internal/protocol/data.go
Termina1: this is a bad idea cause you will have to extend rest/json requests which have nothing to do with selectors

internal/upstreams/flow/execution_flow.go
Termina1: as mentioned above its a bad idea

internal/upstreams/blocks/block_processor.go
Termina1: fixed, also removed unused method

internal/config/option_defaults.go
Termina1: fixed

internal/upstreams/flow/match_trace.go
Termina1: added

KirillPamPam (approved):

PR #205, opened 2026-06-05
Tron rest support

KirillPamPam:
# TRON support

## Summary

Adds first-class TRON support to nodecore. TRON's API surface is shaped differently from every other supported chain — it serves a canonical HTTP REST API on `/wallet/*` (full node) and `/walletsolidity/*` (solidity / confirmed-only mirror) in addition to an Ethereum-compatible JSON-RPC surface on `/jsonrpc`. This PR plumbs all three transports through the existing connector / spec / chain-specific abstractions so operators can configure TRON upstreams with either REST or JSON-RPC, or both.

## Motivation

Before this PR, the only way to route TRON traffic through nodecore was via the curated JSON-RPC subset TRON exposes for Ethereum tooling compatibility. The bulk of TRON's native surface — `triggersmartcontract`, `broadcasttransaction`, `getaccount`, the v2 staking endpoints, witness/governance methods, TRC10 token operations — lives only on `/wallet/*` and was effectively unreachable. Operators of TRON deployments need REST routing to use nodecore at all in production.

### Method specs (data-driven)

`pkg/methods/specs/tron.json` is now a **bundle** that composes three plain sub-specs:

- `tron-json-rpc.json` — 34 methods, the canonical TRON `/jsonrpc` surface (curated `eth_*` / `net_*` / `web3_*` subset, plus TRON's `buildTransaction` extension). Enumerated against [TRON's reference docs](https://tronprotocol.github.io/documentation-en/api/json-rpc/) — no `eth-json-rpc` import, so we don't accidentally inherit methods TRON doesn't actually serve.
- `tron-rest.json` — 84 endpoints, the full `/wallet/*` surface across account / block / transaction / contract / TRC10 / resource-staking (v1 + v2) / witness / governance / proposal / node groups. First standalone use of the `rest` connector type in the repo.
- `tron-rest-solidity.json` — read-only `/walletsolidity/*` mirror declared as `rest-additional` (cannot stand alone as an upstream's primary connector — only pulled in via the `tron` bundle).

Method names follow the `VERB#/path` convention; `cacheable: false` is set on writes and current-state-only reads.

### Chain-specific runtime

`internal/upstreams/chains_specific/tron_specific/tron_rest_specific.go` — implements the full `ChainSpecific` interface for REST-typed upstreams:

- `GetLatestBlock` → `POST /wallet/getblock`
- `GetFinalizedBlock` → parses `solidityBlock` from `/wallet/getnodeinfo`
- `BlockProcessor` reuses `EthLikeBlockProcessor` (poll-based head tracking — WebSocket is out of scope for TRON in this PR)
- `NewTronSpecific` dispatches by connector type: `rest` → `TronRestSpecific`; `json-rpc` → delegates to `evm_specific.NewEvmChainSpecific` so the Ethereum-compat path keeps working

Wired into `upstream_factory.go` via a `chains.IsTron()` branch inside the existing `chains.Ethereum` case (TRON's `chains.yaml` type is `eth`).

### Labels, health, lower bounds

- **Labels** (`internal/upstreams/labels/tron_labels`) — `TronClientLabelsDetector` issues `POST /wallet/getnodeinfo` and reports `configNodeInfo.codeVersion` as the client version, with `"tron"` as the client type.
- **Health validators** (`internal/upstreams/validations/tron_validations`) — `TronPeersValidator` counts `peerList` against `options.MinPeers`. `TronSyncingValidator` projects `(now − blockTime) / 3000` blocks against `chain.Settings.Lags.Syncing`.
- **Lower bounds** (`internal/upstreams/lower_bounds/tron_bounds`) — single binary search over `/wallet/getblock` (empty `{}` body is TRON's "no such block" signal) fans the result out to four `LowerBoundData` entries (`Block`, `State`, `Tx`, `Receipts`), since a TRON node either retains all of these for a given height or none of them.

### Docs

- `docs/nodecore/05-upstream-config.md` — `rest` connector bullet now names TRON alongside Algorand and explains the dual-API model.
- `docs/nodecore/11-method-specs.md` — the bundle section now uses `tron` as the multi-transport example (`tron-json-rpc` + `tron-rest` + `tron-rest-solidity`).

### Misc

`internal/upstreams/upstream_test.go` — 20 tests updated to drain the `InitUpstreamStateEvent` that `BaseUpstream.Start` now emits to announce the upstream's initial state to subscribers.

## Out of scope

- WebSocket subscriptions for TRON (`ParseSubscriptionBlock` / `SubscribeHeadRequest` are intentional stubs).
- `SettingsValidators` for TRON (returns `nil` — chain-id validation against `/wallet/getnodeinfo` is a sensible follow-up if needed).
- DEX/Exchange and Shielded TRC20 REST groups — not part of TRON's canonical HTTP API doc and excluded by design.


tonatoz (changes requested):
## Code review

Found 1 issue:

**1. `/wallet/getblock` is left cacheable, but it returns the chain head when called without `id_or_num`**

In `pkg/methods/specs/tron-rest.json`, both `POST#/wallet/getblock` (L111-114) and `GET#/wallet/getblock` (L115-118) have **no `settings` block**, so they inherit the method default `cacheable: true` (`pkg/methods/method.go`: `cacheable := true`).

TRON's `/wallet/getblock` returns the **latest block** when the request body carries no `id_or_num` (`{}` / `{"detail": false}`) — exactly the form this PR's own `GetLatestBlock` uses (`POST /wallet/getblock` with `{"detail": false}` in `tron_rest_specific.go`). Every other "always latest" block endpoint is explicitly marked `cacheable: false`:

- `getnowblock` (L97-109)
- `getblockbylatestnum`

REST requests return `nil` from `ParseParams`, so there is no latest-tag bypass for them. With an operator-configured cache policy, a `getblock` head request would be served a stale head for the cache TTL — the same hazard already guarded against on `getnowblock` / `getblockbylatestnum`. This looks like an oversight.

Additionally (latent): REST `RequestHash()` hashes only the method string and ignores the request body, so a head request and a specific historical-block request via `getblock` collide on the same cache key.

**Suggested fix:** add `"settings": { "cacheable": false }` to both `getblock` entries.

tonatoz (approved):

PR #209, opened 2026-06-09
Dshackle parity

Termina1:


internal/server/health_server/health_server.go
```diff
@@ -0,0 +1,58 @@
...
+
+	"github.com/drpcorg/nodecore/internal/protocol"
+	"github.com/drpcorg/nodecore/internal/upstreams"
+	"github.com/labstack/echo/v4"
+)
+
+type chainStatus struct {
+	Chain             string `json:"chain"`
+	Status            string `json:"status"`
+	AvailableUpstream bool   `json:"available"`
```
KirillPamPam: Maybe just leave only `Status`? I think there is no reason for a bool value

internal/server/health_server/health_server.go
```diff
@@ -0,0 +1,58 @@
...
+
+type statusResponse struct {
+	Ready  bool          `json:"ready"`
+	Chains []chainStatus `json:"chains"`
+}
+
+func NewHealthServer(supervisor upstreams.UpstreamSupervisor) *echo.Echo {
+	e := echo.New()
+	e.HideBanner = true
+	e.GET("/health", func(c echo.Context) error { return c.NoContent(http.StatusOK) })
```
KirillPamPam, line 24: Do we really need this endpoint? Seems like to be a hardcoded one that always returns 200

internal/server/health_server/health_server.go
```diff
@@ -0,0 +1,58 @@
...
+	Status            string `json:"status"`
+	AvailableUpstream bool   `json:"available"`
+}
+
+type statusResponse struct {
+	Ready  bool          `json:"ready"`
+	Chains []chainStatus `json:"chains"`
+}
+
+func NewHealthServer(supervisor upstreams.UpstreamSupervisor) *echo.Echo {
```
KirillPamPam, line 21: Could you pls add its endpoints and what they are responsible for in the docs?

internal/upstreams/flow/fanout_request_processor.go
```diff
@@ -230,6 +230,10 @@ func (a *maximumValueAggregator) Final(request protocol.RequestHolder) *protocol
 	return totalFailureWrapper(request, protocol.NoAvailableUpstreamsError())
 }
 
+func isNullResponse(response protocol.ResponseHolder) bool {
```
KirillPamPam: Move to not_null_request_processor.go?

internal/upstreams/flow/not_null_request_processor.go
```diff
@@ -0,0 +1,95 @@
...
+		if err != nil {
+			lastErr = err
+			continue
+		}
+		if wrapper == nil || wrapper.Response == nil {
+			lastErr = errors.New("not-null upstream returned empty response")
+			continue
+		}
+		if wrapper.Response.HasStream() {
+			lastErr = errors.New("not-null dispatch received streaming response")
```
KirillPamPam: Stream is a valid result that can be returned to a client

internal/upstreams/flow/not_null_request_processor.go
```diff
@@ -0,0 +1,95 @@
...
+	"github.com/rs/zerolog"
+)
+
+// NotNullRequestProcessor handles methods where a successful JSON-RPC null can
+// be caused by indexing lag on a single upstream rather than by a real network-wide
+// miss. It tries candidate upstreams sequentially and stops on the first successful
+// non-null response. If every upstream returns null, it returns the first null result;
+// otherwise it falls back to the first upstream error response or the last execution
+// error.
+type NotNullRequestProcessor struct {
```
KirillPamPam, line 19: I just realised that such fanout processors should be disabled when `mode` is default since in this mode we tend to reduce number of requests.

internal/upstreams/flow/request_processor.go
```diff
@@ -202,9 +204,138 @@ func sendUnaryRequest(
...
 	}
+	upstreamState := upstream.GetUpstreamState()
+	upstreamNodeVersion := ""
+	if upstreamState.Labels != nil {
+		if version, ok := upstreamState.Labels.GetLabel("client_version"); ok {
+			upstreamNodeVersion = version
+		}
+	}
+
+	requestBlockTag := requestBlockTagMetadata(ctx, request)
```
KirillPamPam: I don't see that this value is used somewhere. It seems to me that we event don't have it at the API level either.

internal/upstreams/flow/request_processor.go
```diff
@@ -202,9 +204,138 @@ func sendUnaryRequest(
...
+	finalizationBlockType, finalizationBlock := responseFinalizationMetadata(upstreamState, requestBlockTag)
+	if lowerBound, ok := liveLowerBoundFromPrunedError(ctx, request, response); ok {
```
KirillPamPam: To be honest I would remove this cause it brings more problems based on the experience with dshackle. It potentially may break everything.

internal/upstreams/labels/static_labels_detector.go
```diff
@@ -0,0 +1,23 @@
+package labels
+
+type StaticLabelsDetector struct {
```
KirillPamPam, line 3: Could you add tests pls

internal/upstreams/flow/not_null_request_processor.go
KirillPamPam: Also it might be great to enable/disable dispatch logic via the config in the `default` mode

KirillPamPam:
In general, could you please add all new logic and config to the docs?

internal/server/health_server/health_server.go
Termina1: its useful to have for k8s deployments

internal/upstreams/flow/request_processor.go
```diff
@@ -319,6 +317,26 @@ func lowerBoundRequestBlock(ctx context.Context, request protocol.RequestHolder)
 	return 0, false
 }
 
+func requestMethodParam(ctx context.Context, request protocol.RequestHolder) specs.MethodParam {
```
KirillPamPam: Why is it necessary to parse it explicitly? The previous version was using `ParseParams` and I think it was ok?

internal/config/option_defaults.go
```diff
@@ -175,6 +175,13 @@ func setOptionsDefaults(
 			lo.Ternary(upstreamMode == StrictMode, true, false),
 		)
 	}
+	if upstreamOptions.EnableNotNullDispatch == nil {
```
KirillPamPam: We also need such params for other dispatch policies

KirillPamPam: Maybe enable/disable such params at the global level but not per upstream? At least per chain.
Otherwise it's unclear what it means - include only enabled upstream in dispatch policies?

internal/upstreams/flow/execution_flow.go
```diff
@@ -325,6 +329,21 @@ func (e *BaseExecutionFlow) createRequestProcessor(request protocol.RequestHolde
...
 
+func (e *BaseExecutionFlow) notNullDispatchEnabled() bool {
+	if e == nil || e.appConfig == nil || e.appConfig.UpstreamConfig == nil {
+		return false
+	}
+	for _, upstream := range e.appConfig.UpstreamConfig.Upstreams {
+		if upstream == nil || upstream.Options == nil || upstream.ChainName != e.chain.String() {
+			continue
+		}
+		if upstream.Options.EnableNotNullDispatch != nil && *upstream.Options.EnableNotNullDispatch {
```
KirillPamPam: Yeah, I think it's better to move such settings at the chain level

internal/upstreams/flow/request_processor.go
Termina1: there was a race and agent fixed it stupidly, I didn't check

Termina1: now its fixed

internal/upstreams/flow/execution_flow.go
Termina1: done

internal/config/upstream_config.go
```diff
@@ -96,8 +113,13 @@ func (u *Upstream) GetBestConnector(upstreamMode UpstreamMode) specs.ApiConnecto
...
 type ChainDefaults struct {
-	PollInterval time.Duration   `yaml:"poll-interval"`
-	Options      *chains.Options `yaml:"options"`
+	PollInterval time.Duration    `yaml:"poll-interval"`
+	Options      *chains.Options  `yaml:"options"`
+	Dispatch     *DispatchOptions `yaml:"dispatch"`
+}
+
+type DispatchOptions struct {
+	NotNull *bool `yaml:"not-null"`
```
KirillPamPam: We need other dispatch policies as well

KirillPamPam (approved):

PR #212, opened 2026-06-11
Aggregate subs + local newHeads

KirillPamPam:
# Subscription aggregation engine + local newHeads

## Why

Today every WebSocket subscription is a blind per-client passthrough: `SubscriptionRequestProcessor`
picks one upstream and sends a real subscribe (`eth_subscribe`, `programSubscribe`, …) to that node
for **each** client. Because the per-upstream WS registry keys on the upstream-returned `subId`
(unique per `Subscribe` call), there is **no cross-client deduplication** — N identical client
subscriptions open N node subscriptions, on every chain.

This PR introduces a per-chain **subscription aggregation engine** (modeled on dshackle's
`EthereumEgressSubscription`) that:

- dedups **all** node-backed subscriptions by `(method + params + selector)` so identical client
  subscriptions share **one** node subscription and fan out from a single channel;
- **synthesizes EVM `newHeads` locally** from nodecore's existing merged-head stream — no node
  subscription at all;
- routes in the flow layer, so **HTTP and gRPC both benefit** automatically.

This is stages 1–2 of the rework. Logs, reorgs and `drpc_pendingTransactions` are intentionally left
for follow-up PRs to keep this one reviewable.

## What's in this PR

**Engine — new package `internal/upstreams/flow/subengine`:**
- `engine.go` — `Engine` interface (concrete `baseEngine`): a ref-counted shared source per
  aggregation key, fan-out via `utils.SubscriptionManager`, teardown shortly after the last
  subscriber leaves, and terminal-error propagation.
- `registry.go` — process-wide per-chain `Registry` (lazy, `utils.CMap`), shared by HTTP and gRPC.
- `heads.go` — `NewHeadsSourceBuilder`: taps `ChainSupervisor.SubscribeState` and emits each new
  block's full header JSON as a `newHeads` notification (no node sub).

**Flow routing (`internal/upstreams/flow`):**
- `sub_processor.go` — all `IsSubscribe()` requests now delegate to the engine. Each client
  generates its own subscription id eagerly; events from the shared source are framed per client.
- `sub_aggregation.go` — the aggregation key (`method | RequestHash | selectorKey`), a deterministic
  `selectorKey`, the generic node-backed source builder (selects an upstream, opens one ws sub,
  swallows the upstream confirmation, surfaces disconnect as a terminal frame), and `newHeads`
  routing (`isNewHeadsRequest` + `localNewHeadsAvailable`).
- `execution_flow.go` — resolves the per-chain engine and injects it.

**Granular subscription capabilities (`internal/protocol`, `internal/upstreams`):**
- `protocol.Cap` gains `NewHeadsCap` / `LogsCap` (EVM) alongside `WsCap` (all chains). They're
  granted in `SubscribeUpstreamStateEvent` only for a **websocket head connector** (so a json-rpc
  head connector with a request-only ws connector does **not** get them) and when the methods expose
  `eth_subscribe` / `eth_getLogs`.
- `ChainSupervisorState.Caps` aggregates caps across available upstreams; `SubMethods` now advertises
  cap-derived topics (`newHeads`, `logs`) for EVM instead of the generic `eth_subscribe`.
- `protocol.Block.RawData` retains the raw header JSON, populated only by EVM
  `ParseSubscriptionBlock`; `HeadWrapper` carries the producing `UpstreamId`.

**gRPC (`internal/server/emerald`):**
- `NativeSubscribe` validates the requested topic against the chain's advertised `SubMethods`
  (`subscribeMethodSupported`) and rejects unsupported topics with `codes.Unimplemented`.

## Behavior

- **Aggregation is per-chain**: one node subscription per `(method+params+selector)` for the whole
  chain, fanned out to all clients.
- **newHeads/logs, two modes**:
  - **HTTP** — local synthesis when the chain has a ws head connector (`NewHeadsCap`), otherwise
    falls back to a real node passthrough.
  - **gRPC** — dshackle-strict, local-only: `newHeads`/`logs` are served only when locally
    synthesizable, otherwise `Unimplemented` (never passed through to a node).
- **No durable reconnect**: on upstream disconnect/error the engine propagates a terminal error to
  all attached clients and tears the shared source down; the client decides whether to resubscribe.

### DIFF
--- a/internal/upstreams/flow/subengine/engine.go
+++ b/internal/upstreams/flow/subengine/engine.go
@@ -4,27 +4,40 @@
 // instead of opening one node subscription per client.
 //
 // The engine is producer-agnostic: callers pass a SourceBuilder that knows how
-// to start the underlying source (a generic node-backed ws subscription, or -
-// in later stages - locally synthesized newHeads/logs). The engine owns
-// ref-counting, fan-out and lifecycle (teardown after the last client leaves).
+// to start the underlying source (a generic node-backed ws subscription, or
+// locally synthesized newHeads/logs).
+//
+// Design: each aggregation key is owned by a single goroutine (a sourceActor).
+// That goroutine holds ALL of the key's mutable state - subscribers, ref count,
+// teardown timer, terminal cause - as ordinary local variables, and is the only
+// thing that ever touches them. There are therefore no mutexes: concurrency is
+// resolved by channel sends to the actor, not by locking shared state. The
+// process-wide key -> actor map is the lock-free utils.CMap.
+//
+// Terminal state is delivered out of band: a subscriber learns the source ended
+// only by its Events channel being closed, after which Subscription.Err returns
+// the cause. The channel never carries an in-band error frame, so a terminal
+// signal can never be lost to a full buffer or missed by a late subscriber.
 package subengine
 
 import (
 	"context"
-	"fmt"
-	"sync"
 	"time"
 
 	"github.com/drpcorg/nodecore/internal/protocol"
-	"github.com/drpcorg/nodecore/internal/upstreams"
 	"github.com/drpcorg/nodecore/pkg/chains"
 	"github.com/drpcorg/nodecore/pkg/utils"
+	"github.com/rs/zerolog/log"
 )
 
 // defaultTeardownDelay mirrors dshackle's refCount(1, 60s): the shared source is
 // kept alive for this long after the last subscriber leaves, so a quick
 // resubscribe reuses it instead of re-opening a node subscription.
 const defaultTeardownDelay = 10 * time.Second
+
+// subscriberBufferSize bounds how far a single subscriber may lag behind the
+// shared source before it is disconnected (see the fan-out in run).
+const subscriberBufferSize = 100
 
 // Source is a started, normalized subscription stream that the engine fans out.
 // Events carries upstream messages plus, on disconnect/error, a terminal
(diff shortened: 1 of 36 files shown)


internal/upstreams/flow/subengine/engine.go
```diff
@@ -0,0 +1,191 @@
...
+		})
+	}
+	return sub.Events, unsub, nil
+}
+
+// pump forwards every source event to the source's subscribers and, when the
+// source ends, removes it from the engine and releases it.
+func (e *baseEngine) pump(key string, s *sharedSource, events <-chan *protocol.WsResponse) {
+	for ev := range events {
+		s.sm.Publish(ev)
```
a10zn8: s.sm.Publish could drop events, maybe we need smth like new publish without droping...

internal/upstreams/flow/subengine/heads.go
```diff
@@ -0,0 +1,65 @@
...
+				case <-srcCtx.Done():
+					return
+				case event, ok := <-sub.Events:
+					if !ok {
+						return
+					}
+					for _, wrapper := range event.Wrappers {
+						head, ok := wrapper.(*upstreams.HeadWrapper)
+						if !ok || len(head.Head.RawData) == 0 {
+							continue
```
a10zn8: Could it lead to silent but live connection?

a10zn8 (approved):

a10zn8 (approved):

PR #216, opened 2026-06-17
Local logs subscription

KirillPamPam:
# Local EVM `logs` subscriptions + reorg handling (subscription rework, stage 3)

## Why

Stages 1–2 of the subscription rework are merged: a per-chain **aggregation engine**
(`internal/upstreams/flow/subengine`) dedups node-backed subscriptions, and EVM `newHeads` is
synthesized locally from nodecore's merged-head stream.

This PR is **stage 3**: synthesize EVM `logs` locally too. Instead of opening
`eth_subscribe("logs")` on a node for every client, the chain watches new blocks, calls
`eth_getLogs {blockHash}` once per block, and fans the logs out to all `logs` subscribers — with
per-client address/topic filtering and reorg-aware `removed:true` re-emits. As with newHeads, routing
lives in the flow layer, so **HTTP and gRPC both benefit**. This mirrors dshackle's
`ConnectBlockUpdates` / `EthereumDirectReader`.

## What's in this PR

**Block-update tracking — `subengine/blockupdates.go` (new):**
- `blockTracker` turns the merged-head stream into ordered `NEW`/`DROP` `BlockUpdate`s via a bounded
  height-indexed history ring. Classification is **by hash per height**, not height alone, so a
  *benign rollback* (a faster upstream dropping out, leaving a lower head from a slower upstream on
  the same chain) is a no-op, while a head that disagrees on the hash at an already-announced height
  (or a forward successor whose parent ≠ our tip) is a **reorg** that drops the orphaned suffix
  `[reorgFrom..tip]`.
- `StreamBlockUpdates` is the thin pump over `ChainSupervisor.SubscribeState`.

**Logs source — `flow/logs_source.go` (new):**
- `newLogsSourceBuilder` builds the chain's single shared **all-logs** source: for each `NEW` block
  it issues one `eth_getLogs {blockHash}` (no address/topic filter), caches the result (FIFO ~32
  blocks), and emits every log as its own event; on a `DROP` it re-emits the cached logs with
  `removed:true`. It terminates (so clients fail over to the generic node-backed path) when the chain
  loses `LogsCap`.
- **Upstream selection is by block HEIGHT, not by the head producer** (dshackle `EthereumDirectReader`
  parity + robustness): a fresh `NewRatingStrategy(chain, "eth_getLogs", [NewHeightMatcher(height)])`
  per block picks any available, best-rated upstream at ≥ that height, with a bounded walk-down on
  error. The head producer may be gone by the time we fetch; height matching tolerates that. We
  already have the height from the head wrapper, so no blockHash→height cache is needed.

**Selection-only execution — `flow/request_processor.go`:**
- `selectAndSend` runs `strategy.SelectUpstream` + the existing `sendUnaryRequest` **without** the
  failsafe executor (no retry/hedge) — the lightweight one-shot path the logs source needs.

**Per-client filtering — `flow/log_filter.go` (new):**
- Exported `SubFilter` interface (`Matches([]byte) bool`) so any shared local source can supply
  per-client filtering; `logFilter` implements it (address set + positional topic matcher, EVM
  semantics). The shared source carries every log; each client drops the ones it didn't subscribe to.

**Routing + wiring:**
- `sub_aggregation.go` — `isLogsRequest`, `localLogsAvailable` (checks `LogsCap`), the per-chain
  `localLogsKey`, and `resolveSource` now returns a `SubFilter` and takes the rating registry. The
  local logs path is gated on `len(request.Selectors()) == 0` so a selector-pinned `logs` sub takes
  the generic node-backed path instead of being silently mis-routed.
- `sub_processor.go` / `execution_flow.go` — the `SubscriptionRequestProcessor` carries the rating
  registry and applies the per-client `SubFilter` in its event loop (newHeads/generic unchanged).
- `subengine/engine.go` — `Source` gains an optional `Buffer` field; the logs source requests a 4096
  per-subscriber fan-out buffer (a busy block can yield thousands of logs) while newHeads/generic keep
  the default 100.

(`protocol.LogsCap` already exists, is granted for a ws head connector exposing `eth_getLogs`, and is
already advertised as the `logs` topic by `processSubMethods` — no changes needed there.)

## Behavior

- **One node-side cost per chain**: a single `eth_getLogs` per block serves all `logs` subscribers,
  regardless of how many distinct address/topic filters they use.
- **Reorgs** re-emit the orphaned block's logs with `removed:true`, then the new block's logs.
- **No durable reconnect**: on `LogsCap` loss the source emits a terminal frame and clients resubscribe
  (they then fall back to the generic node-backed path).

### Accepted limitations (height-only merged head fork-choice)

`HeightForkChoice` only republishes when the max height changes, so:
- a same-height 1-block reorg with no height movement is invisible (it self-heals on the next height
  change via the parent/hash mismatch);
- forward gaps (height jumps) are not backfilled;
- reorg reconciliation is bounded by the history window (~18 blocks).

### DIFF
--- a/internal/upstreams/flow/log_filter.go
+++ b/internal/upstreams/flow/log_filter.go
@@ -1,0 +1,143 @@
+package flow
+
+import (
+	"encoding/json"
+	"strings"
+
+	"github.com/bytedance/sonic"
+	"github.com/drpcorg/nodecore/internal/protocol"
+)
+
+// SubFilter decides, per client, whether a fanned-out event should be delivered
+// to that client. A shared local source (e.g. the chain's single all-logs
+// stream) carries every event for all of its subscribers; each client's
+// SubFilter drops the ones it did not subscribe to. resolveSource returns nil
+// for sources that need no per-client filtering (newHeads, generic passthrough).
+type SubFilter interface {
+	Matches(message []byte) bool
+}
+
+// topicFilter is one position of an eth logs topic filter: any (the request had
+// null at this position) matches anything; otherwise the log's topic at this
+// position must be one of set.
+type topicFilter struct {
+	any bool
+	set map[string]struct{}
+}
+
+// logFilter is a parsed eth_subscribe("logs", {...}) filter. The shared all-logs
+// source carries every log of every block; each client applies its own logFilter
+// in the processor so it sees only the logs it subscribed to. All hex values are
+// compared lowercased.
+type logFilter struct {
+	addresses map[string]struct{} // empty => match any address
+	topics    []topicFilter       // positional; empty => match any topics
+}
+
+var _ SubFilter = (*logFilter)(nil)
+
+// rawLogFilter mirrors the eth logs filter object. address is a string or an
+// array of strings; each topics entry is null, a string, or an array of strings.
+type rawLogFilter struct {
+	Address json.RawMessage   `json:"address"`
+	Topics  []json.RawMessage `json:"topics"`
+}
+
+// parseLogFilter extracts the filter object (params[1]) from an
+// eth_subscribe("logs", {...}) request. A missing object (`["logs"]`) yields a
+// match-everything filter; a malformed object returns an error so the caller can
+// fall back to the generic node-backed path.
+func parseLogFilter(request protocol.RequestHolder) (*logFilter, error) {
+	body, err := request.Body()
+	if err != nil {
+		return nil, err
+	}
+	filter := &logFilter{addresses: map[string]struct{}{}}
+
+	node, err := sonic.Get(body, "params", 1)
+	if err != nil {
+		return filter, nil // no filter object - match everything
+	}
+	raw, err := node.Raw()
+	if err != nil {
+		return filter, nil
+	}
+	var rf rawLogFilter
+	if err := sonic.UnmarshalString(raw, &rf); err != nil {
+		return nil, err
+	}
+
+	for _, addr := range parseStringOrArray(rf.Address) {
+		filter.addresses[strings.ToLower(addr)] = struct{}{}
+	}
+	for _, topic := range rf.Topics {
+		tf := topicFilter{set: map[string]struct{}{}}
+		trimmed := strings.TrimSpace(string(topic))
+		if len(topic) == 0 || trimmed == "null" {
+			tf.any = true
+		} else if vals := parseStringOrArray(topic); len(vals) > 0 {
+			for _, v := range vals {
+				tf.set[strings.ToLower(v)] = struct{}{}
+			}
+		} else {
+			tf.any = true
+		}
+		filter.topics = append(filter.topics, tf)
+	}
+	return filter, nil
+}
+
+// parseStringOrArray decodes a JSON value that is either a single string or an
+// array of strings into a slice (nil for null/empty/other).
+func parseStringOrArray(raw json.RawMessage) []string {
+	if len(raw) == 0 {
+		return nil
+	}
+	var single string
+	if err := sonic.Unmarshal(raw, &single); err == nil {
+		return []string{single}
+	}
+	var many []string
+	if err := sonic.Unmarshal(raw, &many); err == nil {
+		return many
+	}
+	return nil
+}
+
+// Matches reports whether a raw eth log object satisfies the filter. A log with
+// fewer topics than the filter requires does not match (standard eth semantics).
+func (f *logFilter) Matches(logRaw []byte) bool {
+	if f == nil {
+		return true
+	}
+	if len(f.addresses) > 0 {
+		addrNode, err := sonic.Get(logRaw, "address")
+		if err != nil {
+			return false
+		}
+		addr, err := addrNode.String()
+		if err != nil {
+			return false
+		}
+		if _, ok := f.addresses[strings.ToLower(addr)]; !ok {
+			return false
+		}
+	}
+	for i, tf := range f.topics {
+		if tf.any {
+			continue
+		}
+		topicNode, err := sonic.Get(logRaw, "topics", i)
+		if err != nil {
+			return false // log has fewer topics than the filter requires
+		}
+		topic, err := topicNode.String()
+		if err != nil {
+			return false
+		}
+		if _, ok := tf.set[strings.ToLower(topic)]; !ok {
+			return false
+		}
+	}
+	return true
+}
--- a/internal/upstreams/flow/log_filter.go
+++ b/internal/upstreams/flow/log_filter.go
@@ -2,6 +2,7 @@ package flow
 
 import (
 	"encoding/json"
+	"errors"
 	"strings"
 
 	"github.com/bytedance/sonic"
@@ -13,8 +14,35 @@ import (
 // stream) carries every event for all of its subscribers; each client's
 // SubFilter drops the ones it did not subscribe to. resolveSource returns nil
 // for sources that need no per-client filtering (newHeads, generic passthrough).
+//
+// It receives the source-attached protocol.ParsedEvent (parsed once per event in
+// the source) so it can match without re-parsing the raw JSON for every
+// subscriber.
 type SubFilter interface {
-	Matches(message []byte) bool
+	Matches(event protocol.ParsedEvent) bool
+}
+
+type parsedLog struct {
+	raw     json.RawMessage
+	address string
+	topics  []string
+}
+
+var _ protocol.ParsedEvent = (*parsedLog)(nil)
+
+func (p *parsedLog) Raw() []byte { return p.raw }
+
+func parseLogEvent(raw json.RawMessage) *parsedLog {
+	var rl struct {
+		Address string   `json:"address"`
+		Topics  []string `json:"topics"`
+	}
+	_ = sonic.Unmarshal(raw, &rl) // best-effort; missing fields stay zero
+	pl := &parsedLog{raw: raw, address: strings.ToLower(rl.Address), topics: make([]string, len(rl.Topics))}
+	for i, t := range rl.Topics {
+		pl.topics[i] = strings.ToLower(t)
+	}
+	return pl
 }
 
 // topicFilter is one position of an eth logs topic filter: any (the request had
@@ -67,7 +95,15 @@ func parseLogFilter(request protocol.RequestHolder) (*logFilter, error) {
 		return nil, err
 	}
 
-	for _, addr := range parseStringOrArray(rf.Address) {
+	// A present-but-malformed address would otherwise leave the address set empty
+	// and silently match EVERY address (the firehose). Reject it so the caller
+	// falls back to the generic node-backed path, where the node validates the
+	// filter itself.
+	addrs, ok := parseStringOrArray(rf.Address)
+	if !ok {
+		return nil, errors.New("malformed address in logs filter")
+	}
+	for _, addr := range addrs {
 		filter.addresses[strings.ToLower(addr)] = struct{}{}
 	}
 	for _, topic := range rf.Topics {
@@ -75,7 +111,7 @@ func parseLogFilter(request protocol.RequestHolder) (*logFilter, error) {
 		trimmed := strings.TrimSpace(string(topic))
 		if len(topic) == 0 || trimmed == "null" {
 			tf.any = true
-		} else if vals := parseStringOrArray(topic); len(vals) > 0 {
+		} else if vals, ok := parseStringOrArray(topic); ok && len(vals) > 0 {
 			for _, v := range vals {
 				tf.set[strings.ToLower(v)] = struct{}{}
 			}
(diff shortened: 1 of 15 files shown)


internal/upstreams/flow/logs_source.go
```diff
@@ -0,0 +1,217 @@
...
+					}
+					if logsLost() {
+						out <- &protocol.WsResponse{Error: protocol.WsTotalFailureError()}
+						return
+					}
+					switch update.Kind {
+					case subengine.BlockNew:
+						logs, upstreamId := fetchBlockLogs(srcCtx, supervisor, chain, chainSup, registry, update.Block)
+						if logs == nil {
+							continue // fetch failed/skipped (logged); not terminal
```
tonatoz, line 107: Silent log loss on a skipped block. When `fetchBlockLogs` returns `nil` the block is skipped with `continue`, and the client's `logs` subscription silently misses an entire block's logs — no gap signal, no `removed`, nothing. For subscribers that rely on completeness (token-transfer tracking, accounting), this is a permanently-missed set of events the client can never detect. Please make this a conscious choice: at minimum a metric + warn here, and consider terminating the source (see separate note) so clients fail over to the node-backed path instead of losing data.

internal/upstreams/flow/logs_source.go
```diff
@@ -0,0 +1,217 @@
...
+
+	// Select any available, best-rated upstream whose head is at >= the block's
+	// height. A fresh strategy per block carries the height matcher; repeated
+	// SelectUpstream calls walk down the rating list (selectedUpstreams dedup).
+	strategy := NewRatingStrategy(chain, "eth_getLogs", []Matcher{NewHeightMatcher(int64(block.Height))}, chainSup, registry)
+
+	for attempt := 0; attempt < logsFetchAttempts; attempt++ {
+		resp, err := selectAndSend(ctx, supervisor, request, strategy)
+		if err != nil {
+			return nil, "" // strategy exhausted / no upstream at this height
```
tonatoz: Exhausted-strategy skip is completely silent. Unlike the parse-error path above (which logs a warn), this `return nil, ""` — no upstream at the block height, or every attempt errored — has no log line at all. Skipped blocks are therefore invisible in logs and metrics. Add a warn + counter so operators can see when and how often blocks are dropped.

internal/upstreams/flow/logs_source.go
```diff
@@ -0,0 +1,217 @@
...
+						return
+					}
+					if logsLost() {
+						out <- &protocol.WsResponse{Error: protocol.WsTotalFailureError()}
+						return
+					}
+					switch update.Kind {
+					case subengine.BlockNew:
+						logs, upstreamId := fetchBlockLogs(srcCtx, supervisor, chain, chainSup, registry, update.Block)
+						if logs == nil {
```
tonatoz, line 106: Terminate vs. silently drop. A persistently un-servable block (no upstream at height for a while) is currently dropped silently. Given the completeness expectations of `logs`, consider treating a sustained skip like a `LogsCap` loss: emit the terminal frame so clients resubscribe and fall back to the generic node-backed path, rather than continuing to serve a stream with holes.

internal/upstreams/flow/logs_source.go
```diff
@@ -0,0 +1,217 @@
...
+					if !ok {
+						return
+					}
+					if logsLost() {
+						out <- &protocol.WsResponse{Error: protocol.WsTotalFailureError()}
+						return
+					}
+					switch update.Kind {
+					case subengine.BlockNew:
+						logs, upstreamId := fetchBlockLogs(srcCtx, supervisor, chain, chainSup, registry, update.Block)
```
tonatoz, line 105: Head-of-line blocking. `eth_getLogs` is fetched synchronously per block before the next `BlockUpdate` is processed, so the whole chain's logs stream is serialized on one in-flight request. On fast chains (Arbitrum ~250ms blocks) a slow upstream or large block stalls *all* subscribers; the `updates` buffer (64) only delays the back-pressure, which then propagates to the state subscription. It is bounded by `InternalTimeout`, but a prefetch/pipelined fetch or a dedicated worker would decouple block processing from getLogs latency.

internal/upstreams/flow/subengine/blockupdates.go
```diff
@@ -0,0 +1,198 @@
...
+	t.ring[block.Height%historyRingSize] = ringEntry{
+		populated: true,
+		height:    block.Height,
+		hash:      block.Hash,
+	}
+}
+
+// StreamBlockUpdates taps the chain head stream and pushes ordered BlockUpdates
+// to out until srcCtx is cancelled or the head subscription closes. It owns its
+// blockTracker; the caller owns out (it is not closed here).
```
tonatoz: Possible consumer-goroutine leak. `StreamBlockUpdates` returns without closing `out`. If the head subscription closes (`sub.Events` closed) *without* `srcCtx` being cancelled, the logs-source consumer blocks forever in its `select` (no `srcCtx.Done()`, no more `updates`). A `defer close(out)` here would let the consumer exit deterministically via its `if !ok { return }` branch. Edge case (supervisor shutdown), but cheap to harden.

internal/upstreams/flow/logs_source.go
```diff
@@ -0,0 +1,217 @@
...
+						cache.put(update.Block.Hash.ToHex(), logs)
+						for _, raw := range logs {
+							select {
+							case out <- &protocol.WsResponse{Message: raw, UpstreamId: upstreamId}:
+							case <-srcCtx.Done():
+								return
+							}
+						}
+					case subengine.BlockDrop:
+						cached, ok := cache.get(update.Block.Hash.ToHex())
```
tonatoz, line 130: `removed:true` for logs the client never saw added. Because the source is shared and cache-only (32 blocks), a client that subscribes after block N is emitted but before N reorgs will receive `removed:true` for logs it never received as added. Inherent to the shared model and probably acceptable, but please document the client contract: clients must tolerate unmatched/spurious `removed` events.

internal/upstreams/flow/log_filter.go
```diff
@@ -0,0 +1,143 @@
...
+	var many []string
+	if err := sonic.Unmarshal(raw, &many); err == nil {
+		return many
+	}
+	return nil
+}
+
+// Matches reports whether a raw eth log object satisfies the filter. A log with
+// fewer topics than the filter requires does not match (standard eth semantics).
+func (f *logFilter) Matches(logRaw []byte) bool {
```
tonatoz: Per-subscriber re-parse of every log. `Matches` runs `sonic.Get(address)` / `sonic.Get(topics, i)` for every log in every client's processor goroutine — O(#subscribers × #logs) JSON walks per block. On a busy chain with many subscribers this is real CPU. Consider parsing each log's address+topics once in the source and handing filters a lightweight pre-parsed view, reducing it to O(#logs).

internal/upstreams/flow/logs_source.go
```diff
@@ -0,0 +1,217 @@
...
+		}
+		return logs, resp.UpstreamId
+	}
+	return nil, ""
+}
+
+// setRemovedTrue returns a copy of an eth log object with "removed" set to true,
+// for re-emitting a reorged-out block's logs. On any parse/marshal error it
+// returns the input unchanged.
+func setRemovedTrue(raw json.RawMessage) []byte {
```
tonatoz, line 217: `setRemovedTrue` swallows errors silently. On any parse/marshal failure it returns the input *unchanged*, so a reorg DROP would re-emit a log with `removed:false` (or its original value) and the client would treat it as still-valid. Log at debug/warn on the error paths so a malformed cached log doesn't silently produce a wrong (non-)removal.

internal/upstreams/flow/subengine/blockupdates.go
```diff
@@ -0,0 +1,198 @@
...
+// block reorgs anything we have already announced. An exact height+hash match is
+// handled as a duplicate by the caller before this is reached.
+func (t *blockTracker) reorgPoint(block protocol.Block) (uint64, bool) {
+	h := block.Height
+	if h > t.tipH {
+		// Forward: only a direct successor whose parent disagrees with our tip
+		// reveals a (tip) reorg. Forward gaps can't be reasoned about.
+		if h == t.tipH+1 && len(block.ParentHash) > 0 {
+			if e := t.ring[t.tipH%historyRingSize]; e.populated && e.height == t.tipH && !e.hash.Equals(block.ParentHash) {
+				return t.tipH, true
```
tonatoz, line 148: Forward-gap reorg drops the tip but never emits the replacement block's logs. When a reorg surfaces only via a height advance (tip `3:cc` → `4:dd`, parent `c2`), `advance` emits `DROP(cc@3)` + `NEW(dd@4)`, but the new canonical block at height 3 (`c2@3`) is never announced — so its logs are never delivered. This is in the accepted-limitations list, but for `logs` it is a genuine, client-unrecoverable hole (unlike `newHeads`). Please call it out explicitly in client docs and consider whether a one-block backfill on a tip reorg is feasible.

internal/upstreams/flow/subengine/blockupdates.go
```diff
@@ -0,0 +1,198 @@
...
+	}
+	return 0, false
+}
+
+// dropFrom emits BlockDrop for every announced block in [reorgFrom..tipH],
+// top-down, and clears them. It is bounded to the last historyRingSize heights
+// (older entries have already been evicted from the ring).
+func (t *blockTracker) dropFrom(reorgFrom uint64) []BlockUpdate {
+	lo := reorgFrom
+	if t.tipH >= historyRingSize && lo < t.tipH-historyRingSize+1 {
```
tonatoz, line 166: Reorg deeper than the ring is silently un-reconciled. `dropFrom` clamps the lower bound to the retained window, so orphans older than `historyRingSize` are never emitted as `removed` — a silent incompleteness. Emit a metric/warn when `reorgFrom` is clamped (a reorg hit the window bound) so deep reorgs are at least observable.

internal/upstreams/flow/log_filter.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+	var rf rawLogFilter
+	if err := sonic.UnmarshalString(raw, &rf); err != nil {
+		return nil, err
+	}
+
+	for _, addr := range parseStringOrArray(rf.Address) {
+		filter.addresses[strings.ToLower(addr)] = struct{}{}
+	}
+	for _, topic := range rf.Topics {
```
tonatoz, line 109: Unbounded `topics` length from client input. Nothing caps the number of topic positions; a client can send a huge `topics` array and make its own `Matches` loop expensive per log. EVM logs have ≤4 topics — reject (or truncate) filters with more than 4 positions.

internal/upstreams/flow/log_filter.go
```diff
@@ -0,0 +1,143 @@
...
+	raw, err := node.Raw()
+	if err != nil {
+		return filter, nil
+	}
+	var rf rawLogFilter
+	if err := sonic.UnmarshalString(raw, &rf); err != nil {
+		return nil, err
+	}
+
+	for _, addr := range parseStringOrArray(rf.Address) {
```
tonatoz: Malformed `address` silently becomes match-any. If `address` is present but is neither a string nor an array of strings, `parseStringOrArray` returns nil and the address set stays empty → the filter matches *every* address (the client gets the firehose). For a malformed filter that's arguably worse than rejecting; consider returning an error here (→ generic node-backed path) on an unrecognized `address` shape.

internal/upstreams/flow/log_filter.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+	var single string
+	if err := sonic.Unmarshal(raw, &single); err == nil {
+		return []string{single}
+	}
+	var many []string
+	if err := sonic.Unmarshal(raw, &many); err == nil {
+		return many
+	}
+	return nil
```
tonatoz: `parseStringOrArray` conflates 'absent' with 'malformed'. Returning nil for any non-string/array shape means an odd type silently degrades the filter (see the address note above). Distinguish the malformed case so the caller can fall back rather than silently widening the match.

tonatoz (changes requested):

internal/upstreams/flow/logs_source.go
KirillPamPam: Added logs

internal/upstreams/flow/logs_source.go
KirillPamPam: Added logs

internal/upstreams/flow/logs_source.go
KirillPamPam: The same logic is in dshackle, so far I suggest to leave it as is

internal/upstreams/flow/logs_source.go
KirillPamPam: The same logic as in dshackle. I think it's ok for now

internal/upstreams/flow/subengine/blockupdates.go
KirillPamPam: Fixed

internal/upstreams/flow/logs_source.go
KirillPamPam: Added comments. Then it will be added in the docs that describes the whole local sub section

internal/upstreams/flow/log_filter.go
KirillPamPam: Fixed

internal/upstreams/flow/logs_source.go
KirillPamPam: Added logs

internal/upstreams/flow/subengine/blockupdates.go
KirillPamPam: Will be added in the docs

internal/upstreams/flow/subengine/blockupdates.go
KirillPamPam: Added logs

internal/upstreams/flow/log_filter.go
KirillPamPam: I think it's ok for now

internal/upstreams/flow/log_filter.go
KirillPamPam: Fixed

internal/upstreams/flow/log_filter.go
KirillPamPam: Fixed

tonatoz (approved):

PR #220, opened 2026-06-19
viction: disable finalized/safe block polling for PoSV chains

Termina1:
Add chain options SupportFinalizedBlockTag / SupportSafeBlockTag. When set to false (Viction config), cascades to:
- No finalized/safe block polling via eth_getBlockByNumber
- No caching with finalization-type: finalized
- Finalization lag guarded against underflow

Add e2e smoke test for Viction chain startup and basic requests.


internal/config/option_defaults.go
```diff
@@ -12,6 +12,8 @@ func setOptionsDefaults(
 	chainDefaults *ChainDefaults,
 	globalChainOptions *chains.Options,
 	upstreamMode UpstreamMode,
+	chainSupportFinalizedBlockTag *bool,
```
KirillPamPam: Since we use `globalChainOptions`, `chainSupportFinalizedBlockTag` and  `chainSupportSafeBlockTag` from configuredChain.Settings, maybe we just pass configuredChain.Settings to the func instead of each param?

internal/config/option_defaults.go
```diff
@@ -12,6 +12,8 @@ func setOptionsDefaults(
...
+	chainSupportSafeBlockTag *bool,
```
KirillPamPam: Unused?

KirillPamPam (approved):

PR #226, opened 2026-06-22
Pending TX validator + cap processor

KirillPamPam:
# Pending-tx validator + capability detection pipeline

## Why

`PendingTxCap` — the capability the local pending-tx aggregation source
(`flow/pending_tx_source.go`) keys off to decide an upstream can serve
`eth_subscribe("newPendingTransactions")` — was added **unconditionally** the moment a
ws connector connected (`SubscribeUpstreamStateEvent.ProcessEvent`). But a live ws
connection doesn't mean the node actually exposes a usable mempool. dshackle solves
this with `PendingTransactionValidatorImpl`: on BASE it polls `txpool_content` and only
advertises the capability when `pending + queued >= 1000`; other chains keep it
always-on.

Rather than bolt a one-off validator onto the side, this restructures **all** capability
detection into a per-chain pipeline — a `CapProcessor` aggregating `CapDetector`s,
mirroring how `LowerBoundProcessor` aggregates `LowerBoundDetector`s. Cap logic moves
out of the upstream/`startConnectors` into one place where each chain decides its own
caps, and the txpool-gated `PendingTxCap` on BASE becomes just one detector among them.

### New cap pipeline (`internal/upstreams/caps`)

- **`CapDetector`** — streams the cap set it currently asserts (its `Domain()`),
  re-emitting whenever its inputs change. Each detector owns its cadence (push or poll).
- **`BaseCapProcessor`** — fans in all detectors, keeps the latest snapshot per detector,
  unions them, and publishes the merged set **on change only**.
- **`WsPresenceCapDetector`** — generic; asserts a cap while the ws connector reports a
  live connection. Used for `WsCap` everywhere.

### EVM-specific detectors (`internal/upstreams/caps/evm_caps`)

- **`EvmHeadSubCapDetector`** — `NewHeadsCap` / `LogsCap`, gated on a ws head connector
  plus `eth_subscribe` / `eth_getLogs` support.
- **`EvmPendingTxCapDetector`** — `PendingTxCap`, gating ws-presence on a periodic
  `txpool_content` poll: asserted only while ws is up **and** `pending + queued >= txLimit`.
  The threshold is a constructor parameter (`txLimit`) so other chains can be onboarded
  with their own limit; BASE passes `BaseTxLimit = 1000` (dshackle parity).

### Wiring

- `ChainSpecific` gains `CapDetectors(input)`. EVM composes
  `WsCap + head-sub + (BASE → txpool-validated PendingTxCap | other EVM → ws-presence PendingTxCap)`.
- Non-EVM chains (solana, tron-rest, aztec, algorand) use `DefaultCapDetectors`, which now
  returns **only `WsCap`** — `PendingTxCap` is EVM-only and these chains can't serve it
  (previously set globally for no benefit).
- New `BaseCapEventProcessor` emits a replace-set `CapsUpstreamStateEvent`;
  `SubscribeUpstreamStateEvent` and the per-connector state loop in `startConnectors` are
  removed. The processor is driven by the existing `Resume`/`PartialStop` lifecycle.

### Supporting fix

- `ws_processor` only published `WsConnected` on transition, so a cap detector
  subscribing after the connection came up (the cap processor starts in `Resume`, after
  blocking settings validation) would miss it. `SubscribeWsStates` now seeds
  `WsConnected` to a late subscriber when the connector is already connected.

### DIFF
--- a/internal/upstreams/caps/evm_caps/evm_pending_tx_detector.go
+++ b/internal/upstreams/caps/evm_caps/evm_pending_tx_detector.go
@@ -1,0 +1,169 @@
+package evm_caps
+
+import (
+	"context"
+	"encoding/json"
+	"time"
+
+	"github.com/bytedance/sonic"
+	mapset "github.com/deckarep/golang-set/v2"
+	"github.com/drpcorg/nodecore/internal/protocol"
+	"github.com/drpcorg/nodecore/internal/upstreams/connectors"
+	"github.com/drpcorg/nodecore/pkg/chains"
+	"github.com/drpcorg/nodecore/pkg/utils"
+	"github.com/rs/zerolog/log"
+)
+
+const (
+	// BaseTxLimit is the combined pending+queued tx count at or above which a node's
+	// mempool is considered usable for pending-tx subscriptions. Mirrors dshackle's
+	// BASE_TX_LIMIT.
+	BaseTxLimit = 1000
+	// pendingTxValidationInterval is how often txpool_content is polled while the ws
+	// connector is up. Mempool size changes slowly relative to this; dshackle polls
+	// on a similar minutes-scale cadence.
+	pendingTxValidationInterval = 5 * time.Minute
+)
+
+// EvmPendingTxCapDetector asserts PendingTxCap only while the websocket connector is
+// connected AND the node's txpool_content reports at least BaseTxLimit combined
+// pending+queued transactions. It is built only for chains that actually need the
+// check (BASE mainnet); other chains use the generic ws-presence detector. Mirrors
+// dshackle's PendingTransactionValidatorImpl.
+type EvmPendingTxCapDetector struct {
+	name            string
+	wsConn          connectors.ApiConnector
+	internalConn    connectors.ApiConnector
+	chain           chains.Chain
+	internalTimeout time.Duration
+	// txLimit is the combined pending+queued tx count at or above which the mempool
+	// is considered usable. Passed in per chain so different chains can use different
+	// thresholds.
+	txLimit int
+}
+
+func NewEvmPendingTxCapDetector(
+	name string,
+	wsConn connectors.ApiConnector,
+	internalConn connectors.ApiConnector,
+	chain chains.Chain,
+	internalTimeout time.Duration,
+	txLimit int,
+) *EvmPendingTxCapDetector {
+	return &EvmPendingTxCapDetector{
+		name:            name,
+		wsConn:          wsConn,
+		internalConn:    internalConn,
+		chain:           chain,
+		internalTimeout: internalTimeout,
+		txLimit:         txLimit,
+	}
+}
+
+func (e *EvmPendingTxCapDetector) Domain() []protocol.Cap {
+	return []protocol.Cap{protocol.PendingTxCap}
+}
+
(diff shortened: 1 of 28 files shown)


internal/upstreams/ws/ws_processor.go
```diff
@@ -151,6 +158,14 @@ func (b *BaseWsProcessor) SendWsRequest(ctx context.Context, upstreamRequest pro
 }
 
 func (b *BaseWsProcessor) SubscribeWsStates(name string) *utils.Subscription[protocol.SubscribeConnectorState] {
+	// If the connector is already connected, seed the new subscriber with WsConnected
+	// right away: the initial WsConnected was published once on connect and a late
+	// subscriber (e.g. a cap detector started after the connection came up) would
+	// otherwise never learn the connector is live. When disconnected we subscribe
+	// normally - the next connect publishes WsConnected as usual.
+	if b.connected.Load() {
```
tonatoz: **Residual TOCTOU between `connected.Load()` and subscription registration.**

Between this `connected.Load()` returning `true` and `SubscribeWithInitialState` actually registering the subscription in the map, the single-threaded ws loop can process a disconnect: it publishes `WsDisconnected` to all *already-registered* subscribers (this new one isn't in the map yet, so it misses it) and then this subscriber gets seeded with a stale `WsConnected`. The detector would then believe the connector is up until the next reconnect→disconnect cycle corrects it.

This is a logical race, not a data race (`-race` is clean), and the window is narrow since the cap processor subscribes during `Resume` when the connection is usually already stable — so not a blocker. But please at least document it here, or make the snapshot+subscribe atomic under one lock in `SubscriptionManager` to close the window.

internal/upstreams/caps/evm_caps/evm_pending_tx_detector.go
```diff
@@ -0,0 +1,169 @@
...
+				return
+			case state, ok := <-stateSub.Events:
+				if !ok {
+					return
+				}
+				wsConnected = state == protocol.WsConnected
+				// Poll immediately on connect so the cap appears without waiting a full
+				// interval; drop it at once on disconnect.
+				if wsConnected {
+					txpoolFull = e.txpoolAboveLimit(ctx)
```
tonatoz, line 117: **Blocking `txpool_content` poll inside the state select loop.**

`txpoolAboveLimit` is a synchronous request (bounded by `internalTimeout`) executed directly in the select loop. While it runs, the loop can't process an incoming `WsDisconnected`. If ws drops during the poll, the poll result is applied with `wsConnected` still `true` (from the connect event), so `PendingTxCap` is briefly asserted and only retracted on the next iteration once the disconnect is processed.

It self-corrects, so minor — but consider running the poll off the select loop (or re-checking `wsConnected`/`ctx` after it returns) so state transitions aren't delayed by a slow/timing-out node.

tonatoz (changes requested):
Overall a clean, well-tested change. Two minor concerns left inline before merge — see comments.

internal/upstreams/caps/evm_caps/evm_pending_tx_detector.go
KirillPamPam: It's acceptable - there is a buffered channel + this job is very rare

tonatoz (approved):

PR #233, opened 2026-06-25
fix: sanitize metric label values to prevent panic on invalid UTF-8

insider89:
## Problem

A malformed public request whose URI contains non-UTF-8 bytes (e.g. `GET /\xc0`) is turned into a REST request method string of the form `"<VERB>#<path>"`, i.e. `"GET#/\xc0"`. That method string is then used **verbatim as a Prometheus label value**.

`client_golang` panics when a label value is not valid UTF-8:

```
panic: label value "GET#/\xc0" is not valid UTF-8

prometheus.(*CounterVec).WithLabelValues
  flow.(*BaseExecutionFlow).processRequest.func1
    internal/upstreams/flow/execution_flow.go
created by flow.(*BaseExecutionFlow).processRequest
```

The panic happens inside the detached goroutine spawned in `processRequest`, which has **no `recover()`**, so it tears down the whole process (`exit 2`). Because the REST endpoint is public, any client can send a single malformed request and trigger a crash loop.

## Affected label sites

Every label value derived from `request.Method()`:

- `internal/upstreams/flow/execution_flow.go` — `requestTotalMetric`, `requestErrorsMetric`
- `internal/upstreams/flow/request_processor.go` — `hedgeMetric`
- `internal/dimensions/dims.go` — request total/errors/retries/duration (via the dimension key)

## Fix

Add `utils.SanitizeMetricLabel`, which returns valid UTF-8 unchanged and replaces invalid byte sequences with the Unicode replacement character (`U+FFFD`), and apply it to every label value derived from `request.Method()`.

The change is confined to the **metric boundary** — request routing, spec-method resolution and upstream forwarding all still see the original, unmodified method string, so there is no behavioral change for legitimate traffic.

## Tests

`pkg/utils/strings_test.go`:
- valid input is returned unchanged
- the exact production byte sequence (`GET#/` + `0xC0`) becomes valid UTF-8
- a `CounterVec.WithLabelValues` call with the sanitized value does **not** panic (direct regression for the crash)

`go build ./...` and `go test ./pkg/utils/ ./internal/upstreams/flow/ ./internal/dimensions/` pass.

### DIFF
--- a/internal/dimensions/dims.go
+++ b/internal/dimensions/dims.go
@@ -132,7 +132,9 @@ func newUpstreamDimensionKey(chain chains.Chain, upstreamId, method string) upst
 	return upstreamDimensionKey{
 		chain:      chain,
 		upstreamId: upstreamId,
-		method:     method,
+		// method feeds Prometheus labels (see dims.go metric calls); sanitize so
+		// an invalid-UTF-8 method can never panic client_golang.
+		method: utils.SanitizeMetricLabel(method),
 	}
 }
 
--- a/internal/upstreams/flow/execution_flow.go
+++ b/internal/upstreams/flow/execution_flow.go
@@ -254,7 +254,8 @@ func hasHttpConnector(up upstreams.Upstream, requestType protocol.RequestType) b
 func (e *BaseExecutionFlow) processRequest(ctx context.Context, upstreamStrategy UpstreamStrategy, request protocol.RequestHolder) {
 	go func() {
 		defer e.wg.Done()
-		requestTotalMetric.WithLabelValues(e.chain.String(), request.Method()).Inc()
+		metricMethod := utils.SanitizeMetricLabel(request.Method())
+		requestTotalMetric.WithLabelValues(e.chain.String(), metricMethod).Inc()
 
 		if request.SpecMethod() == nil {
 			response := protocol.NewTotalFailure(request, protocol.NotSupportedMethodError(request.Method()))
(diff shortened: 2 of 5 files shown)


KirillPamPam:
@insider89 Hello! Thanks a lot for flagging this issue, it's a really important one. While reviewing we've figured out that it may impact other parts of nodecore as well, not only metrics. So it requires a more broader and general fix. If you don't mind, we are going to implement it on our side. And this PR can be transformed into an issue, for example. What do you think?

insider89:
> @insider89 Hello! Thanks a lot for flagging this issue, it's a really important one. While reviewing we've figured out that it may impact other parts of nodecore as well, not only metrics. So it requires a more broader and general fix. If you don't mind, we are going to implement it on our side. And this PR can be transformed into an issue, for example. What do you think?

Looks good to me. Thank you for support!

KirillPamPam:
> > @insider89 Hello! Thanks a lot for flagging this issue, it's a really important one. While reviewing we've figured out that it may impact other parts of nodecore as well, not only metrics. So it requires a more broader and general fix. If you don't mind, we are going to implement it on our side. And this PR can be transformed into an issue, for example. What do you think?
> 
> Looks good to me. Thank you for support!

Great! Could you create an issue pls where we will track the progress?

insider89:
> > > @insider89 Hello! Thanks a lot for flagging this issue, it's a really important one. While reviewing we've figured out that it may impact other parts of nodecore as well, not only metrics. So it requires a more broader and general fix. If you don't mind, we are going to implement it on our side. And this PR can be transformed into an issue, for example. What do you think?
> > 
> > 
> > Looks good to me. Thank you for support!
> 
> Great! Could you create an issue pls where we will track the progress?

Done https://github.com/drpcorg/nodecore/issues/308

PR #244, opened 2026-06-30
Reference upstream by id (not URL) in connector error messages

artemrootman:
## Problem

When an upstream request fails at the connection layer (DNS failure, connection refused/reset, TLS error, timeout), `http.Client.Do` returns a `*url.Error` whose `Error()` embeds the **full upstream URL** — both any credential carried in its path/query (e.g. `/v2/<key>`, `?dkey=<key>`) and the host itself. `HttpConnector` wrapped that raw error verbatim into the client-facing `ResponseError`, so the URL could surface to the caller. The ctx-cancelled branches and the "invalid upstream url" client-failure paths leaked it the same way.

## Fix

Tag each `HttpConnector` with the id of the upstream it serves (threaded through the connector factory), and make every connector failure path reference the upstream **by id only**:

- `upstream <id> request failed`
- `unable to read response from upstream <id>`
- `invalid url for upstream <id>`
- ctx-cancelled: `upstream <id>: context canceled`

The full URL is still logged for operators (zerolog), but never returned to the caller.

## Tests

A connector test drives a connection failure and asserts the client-facing error references the upstream id while neither the API key nor the host survives in it.


internal/upstreams/upstream_factory.go
```diff
@@ -138,14 +138,29 @@ func createConnector(
 		}
 		return connectors.NewWsConnector(wsProcessor), nil
 	case specs.RestConnector:
-		return connectors.NewHttpConnector(connectorConfig, specs.RestConnector, torProxyUrl)
+		return newHttpConnectorWithId(connectorConfig, specs.RestConnector, torProxyUrl, upId)
 	case specs.RestAdditional:
-		return connectors.NewHttpConnector(connectorConfig, specs.RestAdditional, torProxyUrl)
+		return newHttpConnectorWithId(connectorConfig, specs.RestAdditional, torProxyUrl, upId)
```
a10zn8: Maybe just add up id to NewHttpConnector ?

a10zn8 (approved):

PR #245, opened 2026-06-30
Resolve client IP via trusted proxies instead of trusting all X-Forwarded-For

artemrootman:
## Problem

`ContextWithIps` unioned **every** `X-Forwarded-For` token into the client IP set, and the `allowed-ips` check (`local`/`drpc` key `PreCheckSetting`) matched if **any** token equaled an allowed IP. A client connecting directly could therefore send `X-Forwarded-For: <allowed-ip>` and bypass a key's `allowed-ips` restriction.

## Fix

New `server.trusted-proxies` setting (CIDRs or bare IPs), parsed and validated once at config load.

**When `trusted-proxies` is set**, a **single** client IP is resolved: the direct peer (`RemoteAddr`) unless it is a trusted proxy, in which case the right-most `X-Forwarded-For` entry that is not itself trusted is used. `X-Forwarded-For` from a directly connected client is ignored, so the header cannot be used to spoof an IP.

**When `trusted-proxies` is empty (the default)**, the previous behavior is kept so that existing proxied deployments are not broken by an upgrade: all `X-Forwarded-For` entries remain candidate client IPs. Note that the `allowed-ips` bypass is still possible in this mode - the hardening is opt-in and requires configuring `trusted-proxies`. This is called out with a warning in the docs.

Also in this PR:

- all values of the `X-Forwarded-For` header are read (`Header.Values`), not just the first one - this applies to the legacy path too
- an unparseable `RemoteAddr` falls back to `127.0.0.1` instead of leaking the raw string into IP checks

## Tests

Table-driven coverage in `pkg/utils`: legacy mode collects every entry across repeated headers; trusted mode ignores a spoofed XFF from an untrusted peer, takes the right-most untrusted hop, ignores attacker-prepended entries, falls back to the peer when all hops are trusted; IPv6 and `RemoteAddr` edge cases; `ParseTrustedProxies` validation. Config-side: `trusted-proxies` parsed once at load and invalid entries rejected at startup.

## Docs

`docs/nodecore/02-server-config.md` - the `trusted-proxies` field and a "Client IP resolution" section covering both modes. `docs/nodecore/03-auth.md` - `settings.allowed-ips` updated to the new logic.


internal/config/server_config.go
```diff
@@ -20,6 +21,11 @@ type ServerConfig struct {
 	PyroscopeConfig *PyroscopeConfig `yaml:"pyroscope-config"`
 	GrpcAuthConfig  *GrpcAuthConfig  `yaml:"grpc-auth"`
 	TorUrl          string           `yaml:"tor-url"`
+	// TrustedProxies lists CIDRs (or bare IPs) of reverse proxies in front of
+	// nodecore. X-Forwarded-For is only honored when the direct peer matches one
+	// of these; otherwise the direct peer is used as the client IP. Empty (the
+	// default) means X-Forwarded-For is never trusted.
+	TrustedProxies []string `yaml:"trusted-proxies"`
```
KirillPamPam, line 30: It's necessary to update the docs with new logic

internal/server/http_server/http_server.go
```diff
@@ -97,6 +97,15 @@ func NewHttpServer(ctx context.Context, appCtx *server_ctx.ApplicationServerCont
 
 	httpGroup := httpServer.Group("/queries/:chain")
 
+	// X-Forwarded-For is only trusted from configured reverse proxies; parsed once
+	// here (the config was already validated at load). On error, fall back to no
+	// trusted proxies, i.e. use the direct peer as the client IP.
+	trustedProxies, err := utils.ParseTrustedProxies(trustedProxiesFromConfig(appCtx))
```
KirillPamPam: I think that it's better only once validate and parse trusted proxies during the app startup and then use these values.

pkg/utils/context.go
```diff
@@ -15,27 +17,95 @@ const (
...
+	return prefixes, nil
+}
+
+// ContextWithIps resolves the single client IP for the request and stores it in
+// the context. The client IP is the direct peer (RemoteAddr) unless that peer is
+// a configured trusted proxy, in which case X-Forwarded-For is consulted: walking
+// right-to-left, the first address that is not itself a trusted proxy is the
+// client. When the peer is untrusted, X-Forwarded-For is ignored entirely, so a
+// client connecting directly cannot spoof its IP by sending the header.
+func ContextWithIps(ctx context.Context, request *http.Request, trustedProxies []netip.Prefix) context.Context {
```
KirillPamPam, line 64: I think it's better to save the current logic as well of no trusted proxies are specified. Otherwise we could break someone's services.

pkg/utils/context.go
```diff
@@ -15,27 +17,95 @@ const (
...
-	if ipValues.IsEmpty() {
-		remoteIP, _, err := net.SplitHostPort(request.RemoteAddr)
-		if err != nil {
-			ipValues.Add("127.0.0.1")
-		} else {
-			ipValues.Add(remoteIP)
+	// The peer is a trusted proxy: take the right-most X-Forwarded-For entry that
+	// is not itself a trusted proxy (the proxy appends the connecting IP on the
+	// right, so trusted hops are skipped from the right).
+	if xff := request.Header.Get("X-Forwarded-For"); xff != "" {
```
KirillPamPam: It takes only the first header value, but we it's better to iterate over all of them + fix it in the old logic as well.

pkg/utils/context.go
```diff
@@ -15,27 +17,95 @@ const (
...
+	return peer
+}
 
-	ctx = context.WithValue(ctx, ipKey, ipValues)
+func remoteIP(remoteAddr string) string {
+	if host, _, err := net.SplitHostPort(remoteAddr); err == nil {
+		return host
+	}
+	if remoteAddr != "" {
+		return remoteAddr
```
KirillPamPam, line 149: Previously any unparseable RemoteAddr fell back to 127.0.0.1; now the raw string is returned. Maybe it's better to validate the address before returing?

pkg/utils/context.go
artemrootman: Done, but with a caveat worth stating explicitly: with the legacy path restored, the `allowed-ips` bypass is still reachable in the default configuration - a client connecting directly can send `X-Forwarded-For: <allowed-ip>` and match. The fix is now opt-in via `server.trusted-proxies`.

Implementation: when `trusted-proxies` is empty, `ContextWithIps` behaves as before (every `X-Forwarded-For` entry becomes a candidate client IP, remote address used only when the header is absent). When it is set, a single client IP is resolved as described in the PR body. Both modes are covered by tests in `pkg/utils/context_test.go` (`TestLegacy_*` / `TestClientIP_*`), and the default-mode risk is called out in the docs.

One intentional difference in the legacy path: entries are now trimmed, so `X-Forwarded-For: "1.2.3.4, 5.6.7.8"` yields `5.6.7.8` rather than `" 5.6.7.8"`. The untrimmed value could never match an `allowed-ips` entry, so this only makes matching work as intended.

internal/server/http_server/http_server.go
artemrootman: Done. `trusted-proxies` is parsed in `ServerConfig.validate()` and stored in `trustedProxyPrefixes`, exposed via `ServerConfig.TrustedProxyPrefixes()`. `NewHttpServer` just reads the parsed prefixes, so there is no second parse and no silent error fallback. Covered by `TestServerConfigTrustedProxiesParsedOnce` and `TestServerConfigWrongTrustedProxyThenError`.

pkg/utils/context.go
artemrootman: Fixed. Added `forwardedForIPs`, which flattens all values of the header via `Header.Values("X-Forwarded-For")` (net/http keeps repeated headers as separate entries, so `Header.Get` saw only the first one) and splits each on commas. Both paths - the trusted-proxy resolution and the legacy collect-everything mode - now use it, so the old logic gets the fix as well. Tests: `TestClientIP_RepeatedXFFHeadersAreOrdered`, `TestLegacy_RepeatedXFFHeadersCollected`.

internal/config/server_config.go
artemrootman: Done. `docs/nodecore/02-server-config.md` now documents the `trusted-proxies` field plus a "Client IP resolution" section describing both modes, with a warning about the spoofing risk in the default mode. `docs/nodecore/03-auth.md` no longer describes the old algorithm under `settings.allowed-ips` and links to that section.

pkg/utils/context.go
artemrootman: Fixed. `remoteIP` now validates with `netip.ParseAddr` and falls back to `127.0.0.1` for anything that is not a parseable IP (including a `host:port` whose host is not an IP), so a raw unparseable string never reaches the IP checks. A bare-IP `RemoteAddr` without a port is still returned as-is. Tests: `TestRemoteAddr_UnparseableDefaultsToLocalhost`, `TestRemoteAddr_NonIPHostDefaultsToLocalhost`, `TestRemoteAddr_BareIP`, `TestRemoteAddr_IPv6`.

pkg/utils/context.go
```diff
@@ -13,29 +15,135 @@ type contextKey string
...
+	if !isTrustedProxy(peer, trustedProxies) {
+		return peer
+	}
+	// The peer is a trusted proxy: take the right-most X-Forwarded-For entry that
+	// is not itself a trusted proxy (the proxy appends the connecting IP on the
+	// right, so trusted hops are skipped from the right).
+	forwarded := forwardedForIPs(request)
+	for i := len(forwarded) - 1; i >= 0; i-- {
+		if !isTrustedProxy(forwarded[i], trustedProxies) {
+			return forwarded[i]
```
KirillPamPam: `isTrustedProxy` returns false when netip.ParseAddr fails, so the rightmost-untrusted walk stops on unparseable tokens instead of skipping them. 9.9.9.9, garbage -> garbage. Maybe skip such parts?

artemrootman: Fixed in da64deb. `clientIP` now parses each hop and skips the ones that are not IPs, so the walk continues left instead of returning the garbage token; if no hop is usable it falls back to the trusted peer. `isTrustedProxy` takes a `netip.Addr` now, so "unparseable" and "not trusted" can no longer be confused at the call site.

`9.9.9.9, garbage` -> `9.9.9.9`, `garbage, unknown` -> the peer. A hop that carries a port (`8.8.8.8:1234`) is skipped by the same check. Tests: `TestClientIP_MalformedHopIsSkipped`, `TestClientIP_MalformedHopWithPortIsSkipped`, `TestClientIP_AllHopsMalformedFallsBackToPeer`.

I left the legacy path (empty `trusted-proxies`) collecting tokens verbatim, malformed ones included - that is what it did before, and a non-IP token can only match an equally malformed `allowed-ips` entry. Say the word if you would rather have it filtered there too.

PR #266, opened 2026-07-11
Fix local chain methods over WebSocket

GhostStar:
## Summary

- handle `eth_chainId` and `net_version` before the WebSocket unsubscribe path
- preserve subscription cancellation behavior for local unsubscribe methods
- cover both HTTP and WebSocket request contexts in the local processor test

## Root cause

Every request received through the WebSocket server has a non-nil subscription context. The local request processor treated every local method with such a context as an unsubscribe request and attempted to read `params[0]`. Standard parameterless methods such as `eth_chainId` and `net_version` therefore failed with `value not exists`.

## Tests

- `go test ./internal/upstreams/flow -run TestLocalRequestProcessor -count=1`
- `go test ./internal/upstreams/flow -count=1`
- `go test ./... -count=1`
- `make build`


KirillPamPam (approved):

KirillPamPam:
Thanks for the fix!

PR #283, opened 2026-07-17
feat: NEAR family support (near, near-testnet)

artemrootman:
## What

Adds the NEAR blockchain family (`BlockchainType = "near"`), covering near mainnet and testnet (betanet gets family support for free via the existing registry entry).

Design doc: `docs/superpowers/specs/2026-07-17-near-support-design.md`.

**Stacked on #282** (bitcoin family) — it reuses the JSON-RPC family template; will retarget to main once #282 merges.

## How

- `near_specific`: poll-based optimistic head (`block {"finality":"optimistic"}`, dshackle parity); `GetFinalizedBlock` fetches the `final` head; nearcore has no subscriptions at all, so head tracking is poll-only.
- `near_validations`: chain-id validation via `status.chain_id` (NEAR chain-ids are network-name strings — no genesis call needed); health = syncing flag + stale-head guard (`latest_block_time` vs expected block time × syncing lag) + optional `network_info` peer check.
- `near_labels`: client `neard`, version from `status.version.version`.
- `near_bounds`: sliding GC-window lower bounds from `status.sync_info.earliest_block_height` (STATE+BLOCK, 3-min period) — NEAR non-archival nodes keep ~5 epochs, the node reports its own boundary.
- Method spec `near`: the 14-method dshackle-parity surface, all `cacheable: false` in v1, `send_tx` broadcast. No translations, no bans.

## Testing

- Unit tests across all new packages.
- Live corpus against production near mainnet + testnet nodes: 42/42 checks identical, byte-identical payloads for all deterministic cases (incl. GC'd-block UNKNOWN_BLOCK and the ~10s unknown-tx TIMEOUT_ERROR passthrough). Details in the design doc.


internal/upstreams/chains_specific/near_specific/near_chain_specific.go
```diff
@@ -0,0 +1,195 @@
...
+			n.connector,
+			near_labels.NewNearClientLabelsDetector(n.configuredChain.Chain),
+			n.internalTimeout,
+		),
+	}
+	return labels.NewBaseLabelsProcessor(n.ctx, n.upstreamId, labelsDetectors, n.labelsDelay)
+}
+
+func (n *NearChainSpecificObject) CapDetectors(input caps.DetectorInput) []caps.CapDetector {
+	return caps.DefaultCapDetectors(n.upstreamId, input.WsConnector)
```
KirillPamPam: Should be nil since there is no WS in near

internal/upstreams/chains_specific/near_specific/near_chain_specific.go
```diff
@@ -0,0 +1,195 @@
...
+		detectors,
+	)
+}
+
+func (n *NearChainSpecificObject) HealthValidators() []validations.Validator[protocol.AvailabilityStatus] {
+	if n.options != nil && *n.options.DisableHealthValidation {
+		return []validations.Validator[protocol.AvailabilityStatus]{}
+	}
+	validatePeers := n.options != nil && n.options.ValidatePeers != nil && *n.options.ValidatePeers
+	return []validations.Validator[protocol.AvailabilityStatus]{
```
KirillPamPam: Each health validation should be separate if there are different method calls

internal/upstreams/chains_specific/near_specific/near_chain_specific.go
```diff
@@ -0,0 +1,195 @@
...
+// for near, ~3 blocks ahead of "final".
+func (n *NearChainSpecificObject) GetLatestBlock(ctx context.Context) (protocol.Block, error) {
+	return n.fetchBlockByFinality(ctx, "optimistic")
+}
+
+// GetFinalizedBlock fetches the "final" head. It is not tracked in a separate
+// poll loop (BlockProcessor is nil), but callers asking explicitly get the
+// correct finalized block instead of an optimistic alias.
+func (n *NearChainSpecificObject) GetFinalizedBlock(ctx context.Context) (protocol.Block, error) {
+	return n.fetchBlockByFinality(ctx, "final")
```
KirillPamPam, line 146: In we can detect finality then we need to use a block processor which is above. There is an existing impl `EthLikeBlockProcessor`, but as I see since different chains have finality it's necessary to rename it from `EthLikeBlockProcessor` to `BaseBlockProcessor` (smth like that)

internal/upstreams/chains_specific/near_specific/near_chain_specific.go
```diff
@@ -0,0 +1,195 @@
...
+	"github.com/drpcorg/nodecore/pkg/blockchain"
+	"github.com/drpcorg/nodecore/pkg/chains"
+)
+
+// errUnsupportedHeadSubscriptions is returned by SubscribeHeadRequest and
+// ParseSubscriptionBlock: nearcore's JSON-RPC is HTTP request/response only,
+// there is no subscription transport at all, so head tracking is poll-only.
+var errUnsupportedHeadSubscriptions = fmt.Errorf("near: head subscriptions are not supported")
+
+type NearChainSpecificObject struct {
```
KirillPamPam, line 29: Also pls add tests for this object

internal/upstreams/lower_bounds/near_bounds/near_lower_bound.go
```diff
@@ -0,0 +1,139 @@
...
+
+func nearBounds(bound int64) []protocol.LowerBoundData {
+	bounds := make([]protocol.LowerBoundData, 0, len(nearEmittedBoundTypes))
+	for _, bt := range nearEmittedBoundTypes {
+		bounds = append(bounds, protocol.NewLowerBoundDataNow(bound, bt))
+	}
+	return bounds
+}
+
+func (n *NearLowerBoundDetector) fetchStatus(ctx context.Context) (*nearStatus, error) {
```
KirillPamPam, line 89: Let's simplify that as we have in dshackle. We just need to call one method, probably add a few retries and that's it. It's unnecessary to have a fallback, local bound, etc.

internal/upstreams/validations/near_validations/near_chain_validator.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+func (n *NearChainValidator) Validate() validations.ValidationSettingResult {
+	status, err := fetchNearStatus(n.connector, n.chain.Chain, n.internalTimeout)
+	if err != nil {
+		log.Error().Err(err).Msgf("failed to fetch near status for upstream '%s'", n.upstreamId)
+		return validations.SettingsError
+	}
+	if status.ChainId == "" {
+		log.Error().Err(errNearEmptyChainId).Msgf("failed to validate the chain of near upstream '%s'", n.upstreamId)
```
KirillPamPam, line 41: I think that if `status.ChainId == ""` then this is a FatalSettingError signal

internal/upstreams/validations/near_validations/near_health_validator.go
```diff
@@ -0,0 +1,140 @@
...
+	return &NearHealthValidator{
+		upstreamId:      upstreamId,
+		connector:       connector,
+		chain:           chain,
+		internalTimeout: internalTimeout,
+		validatePeers:   validatePeers,
+	}
+}
+
+func (n *NearHealthValidator) Validate() protocol.AvailabilityStatus {
```
KirillPamPam: Peers and syncing validations should be separate

internal/upstreams/chains_specific/near_specific/near_chain_specific.go
artemrootman: Done - CapDetectors returns nil now.

internal/upstreams/chains_specific/near_specific/near_chain_specific.go
artemrootman: Done - split into NearSyncingValidator (status) and NearPeersValidator (network_info), one method call per validator.

internal/upstreams/chains_specific/near_specific/near_chain_specific.go
artemrootman: Done - renamed EthLikeBlockProcessor to BaseBlockProcessor and near now returns it from BlockProcessor(): the final head is polled by the processor (safe block detection disabled since near has no safe concept).

internal/upstreams/chains_specific/near_specific/near_chain_specific.go
artemrootman: Done - added near_chain_specific_test.go covering ParseBlock, GetLatestBlock/GetFinalizedBlock finality params, subscriptions and cap detectors.

internal/upstreams/lower_bounds/near_bounds/near_lower_bound.go
artemrootman: Done - now a single status call with 3 retry attempts, no fallback/local bound. On error the detector returns (nil, err) and the processor keeps the last published bound, same as the other detectors.

internal/upstreams/validations/near_validations/near_chain_validator.go
artemrootman: Done - empty chain_id is FatalSettingError now.

internal/upstreams/validations/near_validations/near_health_validator.go
artemrootman: Done - separate validators now, see the thread above.

internal/upstreams/chains_specific/near_specific/near_chain_specific.go
```diff
@@ -0,0 +1,212 @@
...
+		n.configuredChain.AverageRemoveSpeed(),
+		detectors,
+	)
+}
+
+func (n *NearChainSpecificObject) HealthValidators() []validations.Validator[protocol.AvailabilityStatus] {
+	if n.options != nil && *n.options.DisableHealthValidation {
+		return []validations.Validator[protocol.AvailabilityStatus]{}
+	}
+	validators := []validations.Validator[protocol.AvailabilityStatus]{
```
KirillPamPam: There is also `ValidateSyncing` option that should be checked. It's like in `EvmChainSpecificObject`

internal/upstreams/validations/near_validations/near_health_validator.go
```diff
@@ -0,0 +1,102 @@
...
+					headAge,
+				)
+				return protocol.Syncing
+			}
+		}
+	}
+	return protocol.Available
+}
+
+func fetchNearStatus(connector connectors.ApiConnector, chain chains.Chain, timeout time.Duration) (*NearStatus, error) {
```
KirillPamPam: NearSyncingValidator could be used as a receiver to avoid passing params

artemrootman: Done - moved the status fetch onto an embedded nearStatusClient, both the syncing and chain validators call it as a receiver method now.

internal/upstreams/chains_specific/near_specific/near_chain_specific.go
artemrootman: Done - the syncing validator is now gated by ValidateSyncing, same as EvmChainSpecificObject.

KirillPamPam (approved):

PR #285, opened 2026-07-17
feat: Starknet family support (starknet, starknet-sepolia)

artemrootman:
## What

Adds the Starknet blockchain family (`BlockchainType = "starknet"`). Also fixes the current half-registered state where a configured starknet upstream fails at startup with `no method spec with name ''`.

Design doc: `docs/superpowers/specs/2026-07-17-starknet-support-design.md`.

**Stacked on #284** (ripple family); will retarget as the stack merges.

## How

The most standard family so far — plain JSON-RPC 2.0, no envelope quirks, no translations:

- `starknet_specific`: poll-based latest head via `starknet_getBlockWithTxHashes("latest")`; poll-only (spec-v0.8 WS subscriptions are a follow-up).
- `starknet_validations`: **real chain-id validation** via `starknet_chainId` (dshackle hardcodes the response and validates nothing) — the hex-felt values (`SN_MAIN`/`SN_SEPOLIA`) already live in `chains.yaml`; syncing health handles all three result forms (bare false/true and the sync object with hex-or-numeric block fields) with the lag threshold.
- `starknet_labels`: two-probe client identity (`pathfinder_version` → `juno_version` fallback) as a direct `LabelsDetector` implementation.
- `starknet_bounds`: **verified**-archive bound — probes block 1 and emits STATE/BLOCK=1 only on success (dshackle hardcodes 1 blindly, which misroutes against pruned nodes); pruned nodes fall back to UnknownBound.
- Method spec: 25-method dshackle-parity surface, `not-null` dispatch on block/tx reads and the three `add*Transaction` submits — deliberately **not broadcast**: starknet answers identical resubmission with `DUPLICATE_TX=59` (an error, unlike EVM's success echo), so fan-out manufactures errors; this is the very reason dshackle's starknet broadcast is broken ("fix broadcast on starknet" TODO).

## Testing

- Unit tests across all new packages (incl. hex-vs-numeric sync fields, mainnet-vs-sepolia chain-id cross-check, pruned-node bound fallback).
- Live corpus against the production sepolia node: 28/29 identical byte-exact (the one "FAIL" was a wrong test premise — `juno_version` is internal-only, matching dshackle). Mainnet corpus impossible today (no node deployed); `SN_MAIN` covered by unit tests. Details in the design doc.


internal/upstreams/chains_specific/starknet_specific/starknet_chain_specific.go
```diff
@@ -0,0 +1,185 @@
...
+			s.connector,
+			s.configuredChain.Chain,
+			s.internalTimeout,
+		),
+	}
+	return labels.NewBaseLabelsProcessor(s.ctx, s.upstreamId, labelsDetectors, s.labelsDelay)
+}
+
+func (s *StarknetChainSpecificObject) CapDetectors(input caps.DetectorInput) []caps.CapDetector {
+	return caps.DefaultCapDetectors(s.upstreamId, input.WsConnector)
```
KirillPamPam: Should be nil since there is no WS

internal/upstreams/chains_specific/starknet_specific/starknet_chain_specific.go
```diff
@@ -0,0 +1,185 @@
...
+	"github.com/drpcorg/nodecore/pkg/blockchain"
+	"github.com/drpcorg/nodecore/pkg/chains"
+)
+
+// errUnsupportedHeadSubscriptions is returned by SubscribeHeadRequest and
+// ParseSubscriptionBlock: spec-v0.8 starknet subscriptions are WS-only and
+// out of v1 scope, so head tracking is poll-only.
+var errUnsupportedHeadSubscriptions = fmt.Errorf("starknet: head subscriptions are not supported")
+
+type StarknetChainSpecificObject struct {
```
KirillPamPam, line 29: Pls add tests

internal/upstreams/chains_specific/starknet_specific/starknet_chain_specific.go
```diff
@@ -0,0 +1,185 @@
...
+	if err != nil {
+		return protocol.ZeroBlock{}, fmt.Errorf("couldn't parse the starknet latest block: %w", err)
+	}
+	return block, nil
+}
+
+// GetFinalizedBlock delegates to GetLatestBlock; a distinct l1_accepted head
+// is a follow-up no consumer asks for today.
+func (s *StarknetChainSpecificObject) GetFinalizedBlock(ctx context.Context) (protocol.Block, error) {
+	return s.GetLatestBlock(ctx)
```
KirillPamPam: Are we sure that in starknet latest == finalized?

internal/upstreams/lower_bounds/starknet_bounds/starknet_lower_bound.go
```diff
@@ -0,0 +1,155 @@
...
+	return &StarknetLowerBoundDetector{
+		upstreamId:      upstreamId,
+		connector:       connector,
+		chain:           chain,
+		internalTimeout: internalTimeout,
+	}
+}
+
+func (s *StarknetLowerBoundDetector) DetectLowerBound(ctx context.Context) ([]protocol.LowerBoundData, error) {
+	err := s.probeEarlyBlock(ctx)
```
KirillPamPam: Dshackle always pushes 1, but if we're sure that it's necessary to calculate it, let's simplify this logic removing fallback and local bound - just use a binary search.

internal/upstreams/validations/starknet_validations/starknet_chain_validator.go
```diff
@@ -0,0 +1,78 @@
...
+	request, err := protocol.NewInternalUpstreamJsonRpcRequest("starknet_chainId", []any{}, s.chain.Chain)
+	if err != nil {
+		return "", err
+	}
+	response := s.connector.SendRequest(ctx, request)
+	if response.HasError() {
+		return "", response.GetError()
+	}
+	chainId := protocol.ResultAsString(response.ResponseResult())
+	if chainId == "" {
```
KirillPamPam, line 77: If it's empty, it cant' be SettingsError, it's a fatal one since we don't now what we have

internal/upstreams/chains_specific/starknet_specific/starknet_chain_specific.go
artemrootman: Done - CapDetectors returns nil now.

internal/upstreams/chains_specific/starknet_specific/starknet_chain_specific.go
artemrootman: Done - added starknet_chain_specific_test.go covering ParseBlock, latest/l1_accepted tags, subscriptions and cap detectors.

internal/upstreams/chains_specific/starknet_specific/starknet_chain_specific.go
artemrootman: Good catch - no, latest is only ACCEPTED_ON_L2. Verified on juno v0.16.4 (RPC spec 0.10.2): latest was block 12262080 while the l1_accepted tag returned 12259300 (ACCEPTED_ON_L1, ~2.8k blocks behind). GetFinalizedBlock now polls the l1_accepted tag (available since RPC spec 0.9) and the block processor is wired to track it. Pre-0.9 upstreams error on the tag and the processor disables finalized detection for them.

internal/upstreams/lower_bounds/starknet_bounds/starknet_lower_bound.go
artemrootman: Went with dshackle parity: constant 1. juno/pathfinder keep the full block history (no block pruning mode exists), so there is nothing to calculate - removed the probe/fallback logic entirely.

internal/upstreams/validations/starknet_validations/starknet_chain_validator.go
artemrootman: Done - empty chain id is FatalSettingError now, fetch failures stay SettingsError.

artemrootman:
Note: this branch still uses the NewEthLikeBlockProcessor name; the rename to BaseBlockProcessor lands in #283, so whichever PR merges second needs a one-line rebase fix.

internal/upstreams/validations/starknet_validations/starknet_syncing_validator.go
```diff
@@ -0,0 +1,123 @@
...
+	syncStatus, err := s.fetchSyncStatus()
+	if err != nil {
+		log.Error().Err(err).Msgf("starknet upstream '%s' health validation failed", s.upstreamId)
+		return protocol.Unavailable
+	}
+	if syncStatus.syncObject == nil {
+		// some upstreams return a bare boolean instead of the sync object
+		if syncStatus.syncing {
+			log.Warn().Msgf("starknet upstream '%s' is syncing", s.upstreamId)
+			return protocol.Syncing
```
KirillPamPam, line 50: I just realized that it's basically a syncing validator. This means that it's necessary to add in its specific object one more check before adding this validator - `options.ValidateSyncing`. And maybe rename StarknetHealthValidator to StarknetSyncingValidator?

artemrootman: Done - renamed to StarknetSyncingValidator and it's added only when ValidateSyncing is on, same as the other chains.

KirillPamPam (approved):

PR #286, opened 2026-07-17
feat: TON family support (ton, ton-testnet)

artemrootman:
## What

Adds the TON blockchain family (`BlockchainType = "ton"`) with the chain's real two-API split treated as first-class: the toncenter **v2 HTTP API** (`rest` connector, base URL carries `/api/v2`) and the **v3 indexer** (`rest-indexer`, a new plain connector type) — 58 methods total (31 v2 + 27 v3).

Design doc: `docs/superpowers/specs/2026-07-17-ton-support-design.md`; operator docs: `docs/nodecore/05-upstream-config.md` (TON deployment modes).

**Stacked on #285** (starknet family); will retarget as the stack merges.

## Deployment modes — v3 is not "additional"

The v2 window is the liteserver's sliding retention; the v3 window is whatever the indexer was backfilled with — independent in both directions (non-archival node + genesis-deep index, or the reverse), with independent failure modes. Per-upstream accounting can't express two windows, so:

- **Split (recommended)**: a standalone v3 upstream (single `rest-indexer` connector — a plain type, legal on its own; `rest-additional` keeps its never-standalone semantics untouched) with fully independent accounting: v3 head, `gen_utime`-freshness health, `global_id` chain validation, `doc.json` labels, and `first.seqno` lower bounds (opting into `DecreasingBoundDetector`: deeper backfill legally moves the bound down). Method routing between the v2/v3 upstreams is automatic via advertised methods.
- **Combined**: both connectors on one upstream still works — all accounting is computed from the v2 API only, the v3 connector serves methods and nothing else, and a warning is logged at upstream creation saying exactly that.

## Family internals

- `ton_specific`: poll head via masterchain info (BFT-final, latest == finalized), v2/v3 flavored by connector set; EmptyHash parent instead of dshackle's random-string hack.
- Race-free v2 chain validation via the zerostate hashes in `getMasterchainInfo.init` (one call; avoids the tonlib masterchain-ref race that made dshackle disable ton validation entirely; testnet zerostate verified against the public endpoint).
- v2 archival-probe bounds (`lookupBlock seqno=1`, honest UnknownBound on non-archival liteservers, not-in-db clears stale cache).
- v2 error envelope needs no new code: non-2xx structured bodies parse through the existing REST error path and pass to clients byte-exact with status parity.

## Testing

- Unit tests across all packages incl. the standalone-additional config-validation gate (allowed for ton, still rejected elsewhere).
- Live corpus against the production mainnet node: 27/27 identical (v2 modulo the volatile `@extra` field), HTTP-status parity on all error passthroughs, v3 empty-200 semantics.
- Deployment modes validated live: split mode shows two independent accounting profiles on one chain (v3 bounds = index floor, v2 = honest UnknownBound); combined mode fires the warning and accounts via v2. Details in the design doc.


internal/upstreams/chains_specific/ton_specific/ton_chain_specific.go
```diff
@@ -0,0 +1,270 @@
...
+// and independent failure modes, so a standalone v3 upstream accounts via
+// the v3 API; a combined upstream accounts via v2 only.
+type tonApiFlavor int
+
+const (
+	tonV2 tonApiFlavor = iota
+	tonV3
+)
+
+type TonChainSpecificObject struct {
```
KirillPamPam: Let's create separate specific objects for v2 and v3 instead of if branches. And then in `NewTonChainSpecificObject` it's easy to return an object based on the specified connector

internal/upstreams/chains_specific/ton_specific/ton_chain_specific.go
```diff
@@ -0,0 +1,270 @@
...
+			t.connector,
+			detector,
+			t.internalTimeout,
+		),
+	}
+	return labels.NewBaseLabelsProcessor(t.ctx, t.upstreamId, labelsDetectors, t.labelsDelay)
+}
+
+func (t *TonChainSpecificObject) CapDetectors(input caps.DetectorInput) []caps.CapDetector {
+	return caps.DefaultCapDetectors(t.upstreamId, input.WsConnector)
```
KirillPamPam: Should be nil since there is no WS

internal/upstreams/chains_specific/ton_specific/ton_chain_specific.go
```diff
@@ -0,0 +1,270 @@
...
+
+	block, err := t.ParseBlock(response.ResponseResult())
+	if err != nil {
+		return protocol.ZeroBlock{}, fmt.Errorf("couldn't parse the ton masterchain info: %w", err)
+	}
+	return block, nil
+}
+
+func (t *TonChainSpecificObject) GetFinalizedBlock(ctx context.Context) (protocol.Block, error) {
+	return t.GetLatestBlock(ctx)
```
KirillPamPam: Are we sure that latest == finalized?

internal/upstreams/chains_specific/ton_specific/ton_chain_specific.go
```diff
@@ -0,0 +1,270 @@
...
+		),
+	}
+	return labels.NewBaseLabelsProcessor(t.ctx, t.upstreamId, labelsDetectors, t.labelsDelay)
+}
+
+func (t *TonChainSpecificObject) CapDetectors(input caps.DetectorInput) []caps.CapDetector {
+	return caps.DefaultCapDetectors(t.upstreamId, input.WsConnector)
+}
+
+func (t *TonChainSpecificObject) LowerBoundProcessor() lower_bounds.LowerBoundProcessor {
```
KirillPamPam: For now I suggest to omit the lower bound logic for ton as we have in dshackle

internal/upstreams/validations/ton_validations/ton_chain_validator.go
```diff
@@ -0,0 +1,82 @@
...
+		upstreamId:      upstreamId,
+		connector:       connector,
+		chain:           chain,
+		internalTimeout: internalTimeout,
+	}
+}
+
+func (t *TonChainValidator) Validate() validations.ValidationSettingResult {
+	expected, hasMapping := expectedZerostates[t.chain.Chain]
+	if !hasMapping {
```
KirillPamPam, line 53: I think if there is no mapping, then it's Fatal since we don't know what real upstream works

pkg/methods/specs/ton-http-v2.json
```diff
@@ -0,0 +1,45 @@
...
+    "title": "TON HTTP API v2 methods",
+    "version": "1.0.0"
+  },
+  "spec": {
+    "name": "ton-http-v2",
+    "api-connectors": ["rest"],
+    "type": "plain"
+  },
+  "methods": [
+    { "name": "GET#/getAddressInformation", "params": [], "settings": { "cacheable": false } },
```
KirillPamPam: Let's format both specs and make it like plain json with each field on the separate line. It's easier to read and modify it

pkg/methods/data.go
```diff
@@ -232,6 +232,7 @@ const (
 	RestConnector
 	GrpcConnector
 	WebsocketConnector
+	RestIndexer    // a self-contained indexer REST API next to the node API (e.g. the TON v3 indexer); a plain type - may be an upstream's only connector
```
KirillPamPam: Let's add it after RestConnector since the order matters here

internal/upstreams/chains_specific/ton_specific/ton_chain_specific.go
artemrootman: Done - split into TonV2ChainSpecificObject and TonV3ChainSpecificObject with a shared embedded base; NewTonChainSpecificObject is now a factory that returns one of them based on the primary connector type.

internal/upstreams/chains_specific/ton_specific/ton_chain_specific.go
artemrootman: Done - CapDetectors returns nil now.

internal/upstreams/chains_specific/ton_specific/ton_chain_specific.go
artemrootman: Yes, for the masterchain head this holds: TON masterchain blocks are finalized by the validators' BFT consensus (Catchain) at production time, and both getMasterchainInfo (v2) and /api/v3/masterchainInfo only ever return a committed block - there are no reorgs of a published masterchain block. Shard blocks are only final once referenced by a masterchain block, but our head tracking is masterchain-only, same as dshackle.

internal/upstreams/chains_specific/ton_specific/ton_chain_specific.go
artemrootman: Done - removed the ton lower bound detectors entirely, LowerBoundProcessor returns nil (dshackle parity: TonLowerBoundService has an empty detectors list).

internal/upstreams/validations/ton_validations/ton_chain_validator.go
artemrootman: Done - missing zerostate mapping is FatalSettingError now.

pkg/methods/specs/ton-http-v2.json
artemrootman: Done - both specs reformatted, one field per line.

pkg/methods/data.go
artemrootman: Done - RestIndexer sits right after RestConnector now.

artemrootman:
Superseded by #294 - same diff re-landed onto main (this one was stacked on the ripple branch, so the merge target was wrong).

PR #287, opened 2026-07-18
feat: Stellar family support (stellar, stellar-testnet)

artemrootman:
## What

Adds the Stellar blockchain family (`BlockchainType = "stellar"`) served via **stellar-rpc** (Soroban RPC). The first family with no dshackle surface to inherit — stellar was never actually served by dshackle (its include-list entry references a chain dshackle doesn't know); the method spec is simply the complete 12-method stellar-rpc JSON-RPC API.

Design doc: `docs/superpowers/specs/2026-07-18-stellar-support-design.md`.

**Stacked on #286** (ton family). **Depends on drpcorg/public#239** — this branch pins the registry submodule to that commit (stellar chain entries, chain-ids = network passphrases); re-pin to the merge commit once it lands.

## How

The cleanest family in the stack — standard JSON-RPC 2.0, standard error codes, zero translations:

- `stellar_specific`: poll head via `getLatestLedger` (SCP finality — ledgers final on close, no reorgs).
- `stellar_validations`: chain validation by `getNetwork` passphrase (the registry chain-ids are the passphrases verbatim; EqualFold compare since the registry loader lowercases chain-ids globally); health via `getHealth` — the node polices its own >30s head staleness, so unhealthy arrives as a -32603 error (bootstrap message → Syncing, staleness → Unavailable).
- `stellar_labels`: `stellar-rpc` + version truncated at the commit suffix (`getVersionInfo`).
- `stellar_bounds`: retention-window bounds from `getHealth.oldestLedger` as BLOCK+TX — deliberately no STATE bound, stellar-rpc has no historical state at all (`getLedgerEntries` is live-only).
- `sendTransaction` stays default-dispatch: it's async enqueue with core statuses in the *result* (PENDING/DUPLICATE/TRY_AGAIN_LATER/ERROR); broadcast with DUPLICATE-as-success reconciliation is a follow-up.

## Horizon

Also adds **Horizon** support with TON-style deployment modes (per review discussion): the full 51-template REST surface on the plain `rest` connector; split mode = a standalone Horizon upstream with fully independent accounting (head via `GET /ledgers?order=desc&limit=1`, `GET /health` health with `core_synced=false` -> Syncing, root `network_passphrase` chain validation, `horizon_version` labels, `history_elder_ledger` lower bounds with decrease support - the retention reaper moves it up, `db reingest range` backfill moves it down); combined mode = accounting via the primary connector + warning. Horizon SSE streaming is out of v1 scope (plain JSON works untouched). Smoke-validated against public SDF Horizon: 9/9, bounds exactly matching the advertised elder ledger, byte-identical problem+json error passthrough; full corpus against our own Horizon lands once its initial ingest completes.

## Testing

- Unit tests across all new packages.
- Live corpus against production mainnet + testnet nodes: 34 PASS / 0 FAIL / 2 SKIP — data arrays byte-exact, NOT_FOUND and error-in-result passthroughs identical, out-of-window -32600 verified shape-wise (the message embeds live window numbers). Details in the design doc.


artemrootman:
Superseded by #295 - same diff re-landed onto main (this one was stacked on the ton branch, so the merge target was wrong).

PR #289, opened 2026-07-20
Set statuses by lags

KirillPamPam:
# Mark lagging upstreams as SYNCING based on cross-upstream head lag

## Summary

Adds an opt-in `validate-lag` behavior: the chain supervisor now derives each
upstream's availability from how far its head trails the chain head. An upstream
that is otherwise `Available` but lags behind the best head by more than the
chain's `lags.syncing` threshold is marked `Syncing`, so routing deprioritizes it
until it catches up. Previously head lag was only recorded as a metric and never
influenced routing.

## How it works

`BaseChainSupervisor.calculateHeadLags()` already computes each upstream's head
lag (chain head height − upstream head height) on every head update. That number
is now also pushed into the upstream itself:

- The supervisor gains a `validateLag` flag, the chain's `syncingLag` threshold
  (resolved once at construction), and a `getUpstream` accessor. It calls
  `Upstream.UpdateHeadLag(lag)` only when an upstream **crosses** the syncing
  threshold — a `lastOver map[string]bool` tracks each upstream's over/under state
  and is cleared when the upstream is removed. Because the derived status only flips
  at that boundary, this bounds emissions to ~2 per syncing episode instead of one
  per block, keeping the upstream emitter off the per-block hot path.
- `calculateHeadLags` no longer early-returns when the dimension `tracker` is nil —
  lag-based status now runs independently of metrics collection.
- `BaseUpstream` stores the current lag in an `atomic.Int64` and emits a
  `StatusUpstreamStateEvent` carrying a `Lag *int64` discriminator.
- In `processStateEvents`, effective availability is recomputed via
  `protocol.StatusByLag(headLag, baseAvail, chain.Settings.Lags.Syncing)`:
  `Available && protocol.LagExceeds(lag, syncing)` ⇒ `Syncing`, otherwise the
  probe-reported status is kept. `LagExceeds` treats a non-positive threshold
  (`0`/unset) as "not configured" and disables the downgrade, so a chain without a
  `lags.syncing` value never marks upstreams `Syncing`. The probe-reported base
  availability (`baseAvail`) is tracked locally in the event loop, so a later
  catch-up restores the original status. The same `LagExceeds` predicate is the
  single source of truth shared by the supervisor's crossing check.

### Event model changes

`StatusUpstreamStateEvent` now distinguishes two triggers via `Lag *int64`:

- `Lag == nil` — a health-probe `setStatus`; updates the base availability.
- `Lag != nil` — a `setLag`; only asks the loop to re-derive availability from the
  current lag. A pointer is used so a real lag of `0` stays distinguishable from a
  status update.

Because both must always be processed, `Same()` now always returns `false`, and
`ProcessEvent` became a no-op — the mutation and the derived-status logic live in
`processStateEvents`, which owns both the base-availability tracking and the
per-chain syncing threshold.

To read the per-chain `lags.syncing` threshold, `BaseUpstream` now holds a
`*chains.ConfiguredChain` instead of a bare `chains.Chain`.

## Configuration

New per-chain flag under `chain-defaults`:

```yaml
upstream-config:
  chain-defaults:
    ethereum:
      validate-lag: true
```

`UpstreamConfig.ValidateLagFor(chain)` resolves it: an explicit per-chain override
always wins; otherwise it defaults to **enabled in `strict` mode** and **disabled
in `default` mode**, consistent with the other real-time validators.


internal/upstreams/chain_supervisor.go
```diff
@@ -270,17 +287,32 @@ func (b *BaseChainSupervisor) calculateFinalizationLags() {
...
+		if b.validateLag && b.getUpstream != nil {
+			// Push a lag update only when the upstream crosses the syncing
+			// threshold (over -> under or under -> over), not on every per-block
+			// lag change. The derived status only flips at that boundary, so this
+			// bounds emissions to ~2 per syncing episode and keeps the upstream
+			// emitter off the per-block hot path.
+			lag := int64(headLag)
+			over := protocol.LagExceeds(lag, b.syncingLag)
+			if b.lastOver[key] != over {
+				b.lastOver[key] = over
```
tonatoz: Set before up nil check

tonatoz (approved):

PR #291, opened 2026-07-21
[DRPCPROT-2072] add capabilities lowerbound check 

l0gun0v:


internal/upstreams/lower_bounds/evm_bounds/capabilities.go
```diff
@@ -0,0 +1,282 @@
...
+	window := c.resultTtl
+	if c.support == evmCapabilitiesUnsupported {
+		window = c.reprobeInterval
+	}
+	if time.Since(c.lastAttemptAt) < window {
+		return c.cached
+	}
+
+	c.lastAttemptAt = time.Now()
+	c.cached = c.refresh(ctx)
```
a10zn8, line 121: Would be nil in case of any error - maybe we need to do few retries?

a10zn8 (approved):

PR #294, opened 2026-07-22
feat: TON family support (ton, ton-testnet)

artemrootman:
Re-lands #286 onto main: the original PR was stacked on the ripple branch (#284), so its merge target was wrong. This is the same reviewed diff (including all review fixes: v2/v3 split objects, no lower bounds, fatal on unknown zerostate, RestIndexer enum order, reformatted specs), applied onto main with the ripple-only context stripped.

Replaces #286.


docs/nodecore/11-method-specs.md
```diff
@@ -174,6 +174,8 @@ nodecore embeds the specs below (see [`pkg/methods/specs/`](../../pkg/methods/sp
 | `bitcoin` | `bitcoin-json-rpc`, `bitcoin-esplora` |
 | `near` | `near-json-rpc` |
 | `starknet` | `starknet-json-rpc` |
+| `ton` | `ton-http-v2`, `ton-index-v3` |
```
KirillPamPam, line 177: In each doc page there might be duplicates

internal/upstreams/chains_specific/ton_specific/ton_chain_specific.go
```diff
@@ -0,0 +1,132 @@
...
+		internalTimeout: options.InternalTimeout,
+		labelsDelay:     options.ValidationInterval * 5,
+		configuredChain: configuredChain,
+	}
+}
+
+// BlockProcessor is nil: the masterchain head is BFT-final the moment it is
+// published, so there is no separate finalized head to poll.
+func (t *tonBaseChainSpecificObject) BlockProcessor() blocks.BlockProcessor {
+	return nil
```
KirillPamPam: Since we detect finaliztion we need to use the generic block processor

internal/upstreams/validations/ton_validations/ton_v3_health_validator.go
```diff
@@ -0,0 +1,81 @@
...
+
+// The indexer commits masterchain blocks in batches, so its head can sit a
+// few dozen seconds behind real time even when perfectly healthy. TON's
+// ExpectedBlockTime * Lags.Syncing is 1s * 40 = 40s - too tight for that
+// commit cadence - so the freshness threshold gets a floor of one minute.
+const tonV3MinFreshness = time.Minute
+
+var errTonV3NoSeqno = errors.New("ton v3 indexer returned zero masterchain seqno")
+
+type TonV3HealthValidator struct {
```
KirillPamPam: Let's rename it to Syncing validator since we check exactly this

artemrootman: Done - renamed to TonV3SyncingValidator.

internal/upstreams/chains_specific/ton_specific/ton_chain_specific.go
artemrootman: Done - both ton objects now wire the generic BaseBlockProcessor polling the finalized head (== latest, the masterchain is BFT-final; safe detection disabled).

docs/nodecore/11-method-specs.md
artemrootman: Good catch - the merge resolutions duplicated a few table rows (ton spec row in 11-method-specs, three TON rows in 05-upstream-config). Deduped.

KirillPamPam (approved):

PR #310, opened 2026-07-29
Refactoring RequestHolder Method to prevent from panics if it's not a valid UTF-8 string

KirillPamPam:
# Stop a non-UTF-8 method name from killing the process (`RequestMethod`)

## Summary

`RequestHolder.Method()` now returns a `protocol.RequestMethod` value instead of a `string`. It
carries the method name in two forms, derived once when the request is built:

- **`Name()`** — byte-for-byte as the client sent it. Routing, spec lookups, method matching, cache
  keys, rate-limit keys, upstream URLs, logs and client-facing error text all use it.
- **`ValidUTF8Name()`** — `strings.ToValidUTF8(name, "�")`. The only form handed to code that
  requires valid UTF-8.

Supersedes #233, which sanitized at three metric call sites. That closed the crash but fixed the
instances rather than the class: every future metric taking a method label was a fresh chance to
reintroduce it. #233 never landed, so nothing here is a revert.

## Motivation

A method name containing invalid UTF-8 reaches `prometheus.CounterVec.WithLabelValues`, which
validates label values and turns the error into `panic(err)`. There is no `recover()` anywhere in
`internal/` or `pkg/`, and echo's `middleware.Recover()` is not registered — so **one request kills
the process**, i.e. a crash loop on a public endpoint. The production crash was the REST path
`GET#/` plus byte `0xC0`.

Two client-reachable entry vectors, not one:

1. **REST** — `rest_parser.go` builds `req.Method + MethodSeparator + "/" + restPath` from the
   URL-decoded path and uses it as the method name whenever no spec template matches. RFC 3986 §2.5
   leaves the encoding of percent-decoded octets undefined, so `%80` is a legal request that simply
   isn't UTF-8.
2. **JSON-RPC body** — sonic's `ConfigDefault` sets `ValidateString: false` and `sonic.Valid` only
   checks syntax, so `{"method":"eth_\x80"}` decodes into a Go string carrying the invalid byte. No
   unusual URL needed, any chain.

(gRPC `NativeCall` is *not* a vector — protobuf-go rejects invalid UTF-8 in proto3 string fields at
unmarshal time.)

Beyond the panic, the same name reached the stats pipeline: the observer is built with it and
`AddResult` runs even on the no-upstream failure path, so it flowed to `base_stats_service.go` →
`drpc_integration.go` `proto.Marshal`, where `string method = 3` requires valid UTF-8. One bad
request could fail an entire `StatsBatch` for an API key.

### The type (`internal/protocol/request_method.go`)

```go
type RequestMethod struct {
	name          string
	validUTF8Name string
}

func NewRequestMethod(name string) RequestMethod   // validUTF8Name: utils.ToValidUTF8(name)
func (m RequestMethod) Name() string
func (m RequestMethod) ValidUTF8Name() string
```

One constructor, private fields, so the two forms cannot diverge. Comparable, so it still works as a
map key. Each request constructor builds it once and passes the same value to both the request's own
field and the observer. `RequestHolder` has only two implementations and no mocks, so the interface
change is contained.

`pkg/utils/strings.go` gains `ToValidUTF8(s string)` — a one-line wrapper so the replacement
character is defined in one place (the ws fix below needs it too, on a value that is not a method
name). `ToValidUTF8` returns valid input unchanged **without allocating**, so building both forms per
request costs a scan, not a copy, and each invalid *run* collapses to a single `�`.

### The valid form is used at 12 conversion sites

| Site | Failure it prevents |
|---|---|
| `flow/execution_flow.go` — `requestTotalMetric`, `requestErrorsMetric`, `quorumVerificationsMetric` ×2 | `panic` → process death |
| `flow/request_processor.go` — `hedgeMetric` | same |
| `caches/cache_processor.go` — `requestCache` | same |
| `dimensions/dimension_hook.go` — the `GetUpstreamDimensions` call | same, but reached as `key.method` rather than `request.Method()` |
| `ws/registry_commands.go` — the three `jsonRpcWsConnectionsMetric` `subscription` labels | same, from an independent taint source (below) |
| `ratelimiter/budget.go` — the `rate_limit_budget_requests` / `_exceeded` labels | same; unreachable today, kept as defence in depth |
| `stats/base_stats_service.go` — the `StatsKey.Method` assignment | `proto.Marshal` failure dropping a whole `StatsBatch` |

`internal/dimensions` needs no change of its own: `GetUpstreamDimensions` has exactly one non-test
caller and `upstreamDimensionKey.method` is only ever a Prometheus label, so sanitizing at the hook
lets the package keep its `string` API — **and that sanitization is load-bearing**. `AddResult` runs
for every `*UnaryResponse`, including one with `UpstreamId == NoUpstream`, and `responseReceive`
dispatches hooks unconditionally; a JSON-RPC method with an invalid byte resolves a *fallback* spec
method, so it clears the `SpecMethod() == nil` gate, fails `MethodMatcher`, and lands here. Verified
live — three tainted names each produced
`nodecore_upstream_requests_total{...,upstream="NoUpstream"}`. See the cardinality note below.

`ratelimiter.Allow` is the same shape: both a semantic consumer (rule matching, engine keys) and a
label sink. Its two label arguments take the valid form, the rules keep the raw name. This one really
is unreachable today — callers gate on `matched.Type() == SuccessType`, hence `MethodMatcher` against
an explicit method set — so it is defence in depth against that gate changing.

### The observer and results carry the whole value

`RequestObserver.method`, `RequestResult.withMethod`, `UnaryRequestResult.method` and `GetMethod()`
are all `RequestMethod`. Nothing narrows the value along the way; the conversion happens where it is
consumed, in the stats key builder. Converting inside `WithMethod` instead would have worked, but
then the stats code would read as if it were copying a plain name and a reader could not tell which
form they had. `statsdata.StatsKey.Method` and the proto field stay plain strings — no structural
change under `internal/stats` or `internal/integration`.

### The ws `subscription` label is a second, independent taint source

`getSubscription` (`ws/ws_protocol.go`) returns `params[0]`'s raw JSON content for `eth_subscribe`
and the method name otherwise — so the value can be arbitrary client bytes with **no method
involved**, which `RequestMethod` cannot cover. Hence `utils.ToValidUTF8` at the three label sites,
with `subType` itself left raw; that also covers the `sub.subType` struct-field read, the same shape
as the `dims.go` case #233 had to patch. Today this is gated by `registry_commands.go`'s
`c.response.Error == nil` (a real upstream rejects a garbage subscription type), so it is a latent
accident rather than a live crash — three lines make it a guarantee.

### Everything else takes `Name()`

49 of the 59 non-test `Method()` call sites (the other 10: 7 take `ValidUTF8Name()`, 3 are
`RequestOperation.Method()`, still a plain string). That covers the 22 `%s` args, the zerolog
`Str("method", …)`, and the semantic uses — matchers, `RateLimiterBudget.Allow`,
`getMethodTranslator`, `BanMethod`, cache-policy matching, `BuildRestURL`, the keydata
contract/method checks, ws spec lookups. `== "eth_subscribe"`-style comparisons stopped compiling,
which is the intended forcing function.

**There is deliberately no `String()` method.** Every site names the form it wants; nothing is
implicitly sanitized. The trade-off was taken knowingly: `%s` on a struct of two strings is legal Go
and `go vet` accepts it, so a missed site prints `{eth_call eth_call}` rather than failing the build.
That is loudly wrong rather than silently wrong, and `grep -n '\.Method()[,)]'` is the standing
guard. A `String()` returning the safe form would have absorbed those 22 sites for free but would
have made the implicit path silently lossy — the opposite of the property we wanted.

## Notes / impact

- **No behavioural change for valid method names.** Both forms are the same string (the same
  backing array), including non-ASCII like `"юникод"`.
- **Logs and client-facing error bodies keep the raw bytes**, on purpose. Logs stay valid JSON
  anyway: zerolog's encoder replaces invalid sequences with `�` itself. `NotSupportedMethodError` /
  `NoApiConnectorsError` reflect the client's own bytes back, which harms nobody — no crash, no
  dropped batch.
- **Metric series for a tainted request now appear as e.g. `method="GET#/status�"`** instead of
  taking the process down.
- **`""` as the replacement was considered and rejected.** It fabricates a plausible name —
  `eth_` + `0x80` + `call` would sanitize to exactly `eth_call`, counting garbage under a real
  method's series — and a fully-invalid name would become an empty label, which Prometheus treats as
  equivalent to the label being absent.
- **Label cardinality is unchanged and remains a non-goal — but it is wider than the two
  client-facing counters, and worth knowing before this merges.** Because the dimension hook fires
  for no-upstream results, each distinct *unknown* method name also mints, under
  `upstream="NoUpstream"`: one `nodecore_upstream_requests_total` series **plus a full
  `nodecore_upstream_request_duration` histogram** (13 buckets + `_sum` + `_count`) — ~16 series —
  **plus** an entry in `BaseDimensionTracker.upstreamDimensionsMap`, a `LoadOrStoreLazy` map with no
  eviction that holds a quantile tracker per entry. So the exposure is memory as well as series
  count, and it is driven by any unknown method name, valid UTF-8 or not.
  This is **pre-existing** — this PR neither introduces nor worsens it, and it makes those labels
  *valid* where they previously crashed the process. It is called out because the follow-up must have
  dimensions in scope, not just the two counters. One option worth considering there: skip the
  dimension hook entirely for `NoUpstream` results — arguably correct regardless of cardinality,
  since the latency of a request that never left is not an upstream dimension.
- **No lint/AST backstop was added.** A syntactic rule cannot see through a struct field or a helper
  (it would have missed the `dims.go` case), and the type already puts the safe form within reach at
  every sink.
- `pkg/test_utils/matchers.go` needed `cmpopts.EquateComparable(protocol.RequestMethod{})` — go-cmp
  otherwise panics descending into the new struct's unexported fields.

## Docs

`docs/superpowers/specs/2026-07-29-request-method-valid-utf8-design.md` — design doc: both entry
vectors with their root causes, the decisions table, the full sink inventory, and the edge cases
(including why `""` was rejected as the replacement string and what the non-goals cover).

## Testing

87 files, +608/−176. Most of the line count is mechanical: 89 `.Method()` sites across ~53 test
files, overwhelmingly `assert.Equal(t, "eth_call", req.Method())` → `.Name()`.

New tests:

- `pkg/utils/strings_test.go` — valid input (incl. `"юникод"`, emoji) unchanged; every repro
  sequence (`%80`, `%FF`, `%C3`, `%F0%9F%98`, `%ED%A0%80` surrogate, `%C0%AF` overlong, invalid
  inside valid) maps to the expected string and passes `utf8.ValidString`; an invalid run collapses
  to one `�`.
- `internal/protocol/request_method_test.go` — both forms for valid names; `Name()` stays
  byte-identical for an invalid one while `ValidUTF8Name()` is valid; comparability. Plus the
  regression proper: a real `prometheus.CounterVec` accepts the label form **and still panics on the
  raw form** — asserting both directions keeps the reason for `ValidUTF8Name` visible. Plus the
  observer/result carrying both forms through to what the stats key builder consumes.
- `internal/ratelimiter/budget_test.go` — `Allow` with an invalid-UTF-8 method neither panics nor
  changes rule matching (the second call still exhausts the budget, i.e. the rule matched the raw
  name).
- `rest_request_test.go` template expectations unchanged — the guard proving `Name()` is still
  byte-exact for routing and cache keys.

`go build ./...`, `go vet ./...`, `golangci-lint run ./...` (0 issues) and `go test -race -p 8 ./...`
all pass. The two `internal/caches/*_e2e` packages fail in environments without Docker — they spin up
containers in `TestMain` — and fail identically on a clean `origin/main` checkout.

Verified end-to-end against a running instance (a stub EVM node so the chain was genuinely
*available* — otherwise "no available upstreams" short-circuits before the crash site):

- REST `status%80` and `%C0`; JSON-RPC bodies with a raw `0x80`, a UTF-16 surrogate, an overlong
  encoding, and a batch mixing a good and a tainted entry → error responses, **process alive across
  all of them, zero panics**;
- metric labels contain `ef bf bd` (U+FFFD, confirmed by hexdump) while the error bodies contain the
  raw byte — exactly the intended split;
- successful requests (which exercise `AddResult` → observer → result) and normal traffic on both
  chains unaffected.


internal/ratelimiter/budget.go
```diff
@@ -33,7 +34,8 @@ type RateLimitBudget struct {
 }
 
 func (b *RateLimitBudget) Allow(method string) (bool, error) {
-	rateLimitRequestMetrics.WithLabelValues(b.Name, method).Inc()
+	methodLabel := utils.ToValidUTF8(method)
```
tonatoz, line 37: `Allow` derives the label form on every call, and it runs once per candidate upstream inside `processMatchedResponse` / `filterUpstreams`, so the same name is scanned repeatedly. This is the one place where the new type's derive-once property is dropped.

Passing `protocol.RequestMethod` is not an option here: `internal/protocol/data.go:12` already imports `internal/ratelimiter`, so that direction is an import cycle. Suggest `Allow(method, methodLabel string)` instead, with `strategy.go:209` supplying `request.Method().Name()` and `request.Method().ValidUTF8Name()` - the caller already holds both forms.

internal/protocol/request_method.go
```diff
@@ -0,0 +1,37 @@
...
+	}
+}
+
+func (m RequestMethod) Name() string {
+	return m.name
+}
+
+func (m RequestMethod) ValidUTF8Name() string {
+	return m.validUTF8Name
+}
```
tonatoz, line 37: The description rejects `String()` because it would make the implicit path silently lossy. That argument applies to a `String()` returning the *safe* form, but not to one returning the raw name:

```go
func (m RequestMethod) String() string { return m.name }
```

This variant matches the rule already stated in the description (logs and client-facing errors keep the raw bytes), removes the `{eth_call eth_call}` failure mode entirely instead of making it loud, and lets the ~22 format-string sites drop `.Name()`. Nothing is implicitly sanitized, and the forcing function is preserved - `== "eth_subscribe"` still does not compile. Worth reconsidering before the two-form convention is locked in across 59 call sites.

internal/upstreams/flow/execution_flow.go
```diff
@@ -259,10 +259,10 @@ func hasHttpConnector(up upstreams.Upstream, requestType protocol.RequestType) b
 func (e *BaseExecutionFlow) processRequest(ctx context.Context, upstreamStrategy UpstreamStrategy, request protocol.RequestHolder) {
 	go func() {
 		defer e.wg.Done()
-		requestTotalMetric.WithLabelValues(e.chain.String(), request.Method()).Inc()
+		requestTotalMetric.WithLabelValues(e.chain.String(), request.Method().ValidUTF8Name()).Inc()
 
 		if request.SpecMethod() == nil {
-			response := protocol.NewTotalFailure(request, protocol.NotSupportedMethodError(request.Method()))
+			response := protocol.NewTotalFailure(request, protocol.NotSupportedMethodError(request.Method().Name()))
```
tonatoz, line 265: The raw name reaches the client-facing JSON body, and sonic emits invalid bytes verbatim rather than escaping them. Checked on this branch:

```
sonic:  {"m":"eth_\x80"}    -> utf8.ValidString == false
stdlib: {"m":"eth_\ufffd"}  -> utf8.ValidString == true
```

`response.go:575` marshals `jsonRpcErr` through sonic, so the error body is not valid UTF-8 and therefore not valid JSON per RFC 8259 section 8.1. Go clients tolerate it, stricter parsers (serde_json and similar) reject the whole response.

The client already knows the bytes it sent, so `ValidUTF8Name()` loses no information here. Same applies to `NoApiConnectorsError` at `request_processor.go:208` and `sub_aggregation.go:234`.

internal/dimensions/dimension_hook.go
```diff
@@ -19,7 +19,9 @@ func (d *DimensionHook) OnResponseReceived(
 		for _, result := range request.RequestObserver().GetResults() {
 			switch r := result.(type) {
 			case *protocol.UnaryRequestResult:
-				dims := d.tracker.GetUpstreamDimensions(r.GetChain(), r.GetUpstreamId(), request.Method())
+				// The valid-UTF-8 form: the dimension key's method ends up as a
+				// Prometheus label in dims.go, where an invalid one would panic.
+				dims := d.tracker.GetUpstreamDimensions(r.GetChain(), r.GetUpstreamId(), request.Method().ValidUTF8Name())
```
tonatoz, line 24: This sanitization is only required because the hook also runs for `UpstreamId == NoUpstream`, as the description explains. A guard on that condition would achieve more with less:

- a tainted name cannot arrive via a real upstream (`MethodMatcher` rejects an unknown method), so the sanitization here becomes unnecessary rather than load-bearing;
- it removes the unbounded growth of `BaseDimensionTracker.upstreamDimensionsMap` (no eviction, one quantile tracker per entry) plus the ~16 series per distinct unknown method name that the description lists as a follow-up;
- it is correct independently of cardinality - the latency of a request that never left is not an upstream dimension, which is the argument the description already makes.

`NoUpstream` currently lives in `flow` (`strategy.go:15`) and `flow -> upstreams -> dimensions`, so the constant would need to move to `protocol`, where it is already the value assigned to `ResponseHolderWrapper.UpstreamId`. That is a small change, and it closes the taint path instead of covering it.

internal/upstreams/ws/registry_commands.go
```diff
@@ -76,7 +77,9 @@ func (c *rpcCommand) handle(registry *BaseRequestRegistry) {
 					ops:     make(map[string]RequestOperation),
 				}
 				state.subs[subID] = sub
-				jsonRpcWsConnectionsMetric.WithLabelValues(registry.chain.String(), registry.upId, req.SubType()).Inc()
+				// The sub type is client-controlled - for eth_subscribe it is the
+				// raw params[0] - so it needs the valid-UTF-8 form as a label.
+				jsonRpcWsConnectionsMetric.WithLabelValues(registry.chain.String(), registry.upId, utils.ToValidUTF8(req.SubType())).Inc()
```
tonatoz, line 82: `registrySubscription.subType` stays raw and is sanitized at three separate label sites (here, `:146`, `:167`). It is genuinely needed raw for `requestOpBufferSize` (`data.go:249`), so the same reasoning behind `RequestMethod` applies: store the label form once on `registrySubscription` next to the raw one and have the three `WithLabelValues` calls read it. It also makes the `Inc`/`Dec` pair read from the same field rather than from `req.SubType()` on one side and `sub.subType` on the other.

internal/protocol/request_method_test.go
```diff
@@ -0,0 +1,84 @@
...
+	)
+
+	for _, request := range []protocol.RequestHolder{restRequest, jsonRpcRequest} {
+		assert.NotPanics(t, func() {
+			metric.WithLabelValues("cosmos", request.Method().ValidUTF8Name()).Inc()
+		})
+		// The raw form is what used to reach the label and crash the process.
+		// Asserting it still panics keeps the reason for ValidUTF8Name visible.
+		assert.Panics(t, func() {
+			metric.WithLabelValues("cosmos", request.Method().Name()).Inc()
```
tonatoz, line 55: This assertion pins behaviour of a dependency rather than of this package, and will fail when client_golang changes its label validation (there is ongoing upstream work on UTF-8 name and label schemes).

The `NotPanics` assertion above already covers the contract under test. `assert.False(t, utf8.ValidString(request.Method().Name()))` documents the same reason for `ValidUTF8Name` without depending on a panic in a third-party library.

tonatoz (changes requested):

internal/ratelimiter/budget.go
KirillPamPam: `Allow`'s param is an abstract identifier, so I didn't want to change its signature. It's easier to have it inside this specific impl.

internal/protocol/request_method.go
KirillPamPam: I deliberately didn't want to have `String()` since it may confuse and it's easy to call it in metrics and get a panic, I wanted to have 2 explicit methods

PR #312, opened 2026-07-30
Reject non-utf8 methods

KirillPamPam:
# Reject non-UTF-8 method names at the parser

## Why

A single request with a raw invalid UTF-8 byte in `"method"` kills the nodecore process.

`prometheus/client_golang` validates every label value the moment it is set — `validateLabelValues`
runs an unconditional `utf8.ValidString` (`prometheus/labels.go:177-181`) — and `WithLabelValues`
turns the failure into a **panic** rather than a returned error (`counter.go:283-289`,
`gauge.go:237-243`). So the exposition is never corrupted; a bad value simply cannot enter the
registry.

The problem is where that panic lands. `internal/upstreams/flow/execution_flow.go:259-262`:

```go
func (e *BaseExecutionFlow) processRequest(...) {
	go func() {
		defer e.wg.Done()
		requestTotalMetric.WithLabelValues(e.chain.String(), request.Method()).Inc()
```

The metric call is the first statement of a **detached goroutine**, before the
`SpecMethod() == nil` check — so an unknown or garbage method name is not filtered out first. There
is no `recover()` anywhere in `internal/`, `pkg/`, or `cmd/`, and `net/http`'s `conn.serve` recover
only protects the handler's own goroutine, not this one. In Go an unrecovered panic in any goroutine
terminates the whole program, so the process dies — every in-flight request across every chain, and
every upstream connection with it.

`sonic` does not sanitize string fields, so the byte arrives intact: `{"method":"eth_\xffcall"}`
decodes with `err == nil` and `utf8.ValidString(Method) == false`. Verified, along with everything
else asserted here.

There is a second instance of the same class at `internal/upstreams/ws/registry_commands.go:79`,
which labels with the `eth_subscribe` subscription type.

Reaching either requires passing auth (`http_server.go:118` runs before `handleHttp` at `:137`), so
this is a valid-credential DoS rather than an open-internet one.

## What changes

Reject a request **if and only if its method name is not valid UTF-8**, at the three places a
client-controlled string becomes a method name or a subscription label:

| Site | Change |
| --- | --- |
| `NewJsonRpcHandler` (`handlers.go`) | New `errNonUtf8Method` sentinel; loop over the decoded batch rejecting any entry whose `Method` fails `utf8.ValidString`. In the constructor, not `RequestDecode`, so the HTTP entry (`http_server.go:189`) and the WS entry (`ws_server.go:100`) are both covered by one check, before any work is scheduled. |
| `parseRestRequest` (`rest_parser.go`) | One check on the resulting `methodTemplate`, covering both switch branches. The matched-spec-template branch passes trivially (templates come from the embedded spec JSON); only the `"<VERB>#/<restPath>"` fallback can reject. |
| `getSubscription` (`ws_protocol.go`) | Returns `(string, error)`; validates the `eth_subscribe` subscription type. `RequestFrame` propagates in the existing `fmt.Errorf("... cause - %s")` style, so the error surfaces from `sendWsRequest` as an ordinary per-request upstream error — the client's connection stays open and no metric series is created. |

Because each value is validated at its source, nothing downstream needs a raw/sanitized split.
`RequestOperation`, `BaseRequestOp`, `registrySubscription`, and `request_registry.go` have **zero
diff**, and every existing `method` label site (`caches`, `dimensions`, `ratelimiter`, `rating`,
`flow`) is covered without being touched.

## Two latent bugs fixed in passing

Both live in the expression `getSubscription` had to modify, and both are reachable from a client:

- `{"method":"eth_subscribe","params":[1]}` **panicked**. `Raw()` on a number node returns the
  1-character string `"1"`, so `sub[1:len(sub)-1]` was `sub[1:0]` → `slice bounds out of range
  [1:0]`.
- `{"method":"eth_subscribe","params":[{"a":1}]}` silently produced the label `"a":1`.

A `TypeSafe() == ast.V_STRING` guard (the repo's existing idiom — `integrity_methods.go:144`,
`keydata/data.go:51`, `parse_response.go:40`) plus a defensive `len(sub) >= 2` closes both. A
non-string first param is *not* a rejection reason: it falls through to the method name and the
upstream decides.

## Deliberate decisions

- **Only the method name rejects.** Invalid bytes in params, headers, query values, or REST wildcard
  captures pass through untouched — they never become a method name or a label. Three negative tests
  pin this, because over-rejection would be as much a defect as under-rejection.
- **A batch fails whole.** One bad method rejects the entire array via `protocol.ParseError()`, which
  is how every other parse failure already behaves. Per-entry rejection would need response-holder
  machinery in `RequestDecode`.
- **On WebSocket the connection closes.** `ws_server.go:100-104` breaks the read loop on any
  handler-construction error — pre-existing behavior, unchanged.
- **Escaped input is accepted.** `Raw()` returns escaped source text, so a body containing
  `"\ud800"` yields the six ASCII characters `\ud800`, which is valid UTF-8 and cannot panic
  `WithLabelValues`. Only genuinely raw invalid bytes are rejected — exactly the set that matters.

## Cost

`utf8.ValidString` on a 20-byte method name: **3.9 ns**, zero allocations. 1 KB: 14.9 ns. A
pathological 1 MB method name: 11 µs — bounded by work already done, since sonic had to parse and
allocate that same payload to hand it over. Invalid input is the cheapest path (1.7 ns for a bad
first byte), so the check cannot be used as an amplifier. Across a batch the total is O(body), a
small constant factor on top of the parse that just ran. No length cap or fast path needed.

## Testing

12 new tests. `make lint`: 0 issues. Full suite passes except `internal/caches/{postgres,redis}_e2e`,
which require Docker and fail only because it was not running locally; neither package references the
changed code.

- `handlers_test.go` — invalid method rejected (single and batch); well-formed multi-byte UTF-8
  accepted; invalid bytes in `params` accepted. The rejection tests assert the sentinel's exact
  message, so they cannot silently pass if a future sonic version starts rejecting invalid UTF-8
  itself.
- `rest_parser_test.go` — invalid `restPath` on the fallback branch rejected (`require.ErrorIs`);
  matched template with invalid bytes in a wildcard capture accepted; invalid bytes in a query value
  or header accepted.
- `ws_protocol_test.go` — invalid subscription type rejected; `params:[1]` and `params:[{"a":1}]`
  fall back to the method name with no panic; valid subscription type unchanged. The three
  pre-existing `RequestFrame` tests still pass, pinning that a non-`eth_subscribe` subscribe method
  returns the method name and a non-subscribe request returns `""`.

## Out of scope

- **The `chain` path parameter as a label.** `ws_server.go:58,61` uses the raw `:chain` value, so
  `GET /queries/%FFfoo` with a WS upgrade still panics `WithLabelValues` — but on the handler's own
  goroutine, where `conn.serve`'s recover catches it: the connection drops, the process survives, and
  the deferred `Dec` never registers so there is no gauge leak. Not a method name, so outside this
  change's rule. Worth a follow-up.
- **No `recover()` in the detached goroutine** at `execution_flow.go:259`. This change removes the
  known trigger; the fragility remains, and any future panic there still kills the process. A single
  `recover()` would convert a whole class of process-death bugs into a failed request. Worth a
  follow-up, ranked above the item above.
- **Subscription label cardinality.** Any distinct *valid* subscription type already creates its own
  series. Pre-existing and unrelated to UTF-8 validity.
- **The gRPC entry point** passes `item.GetMethod()` in unchecked, but `blockchain.proto` is proto3
  and protobuf's proto3 string codec validates UTF-8 on the wire, so a malformed name is rejected
  before nodecore sees it. Note that `NativeSubscribe`'s params arrive as an unvalidated `bytes`
  field, so that path depends on the new `getSubscription` check.
- Other places a method name reaches a log field. `zerolog`'s JSON encoder replaces invalid bytes
  with U+FFFD, so a log field cannot emit invalid UTF-8 independently of this change.

### DIFF
--- a/docs/superpowers/specs/2026-07-30-reject-non-utf8-methods-design.md
+++ b/docs/superpowers/specs/2026-07-30-reject-non-utf8-methods-design.md
@@ -35,9 +35,16 @@ Two pre-existing defects live in the same expression that has to be touched
 
 ## Rule
 
-Reject a request if and only if its **method name** is not valid UTF-8. Nothing else is a
-reason to reject: invalid bytes in params, headers, query values, or REST wildcard captures
-pass through untouched, because they never become a method name or a metric label.
+Reject a request if and only if a value that becomes a metric label is not valid UTF-8. In
+practice that is the **method name** plus the **`chain` path parameter** (added in review — see
+section 4). Nothing else is a reason to reject: invalid bytes in params, headers, query values,
+or REST wildcard captures pass through untouched, because they never become a label value.
+
+On the REST branch this rejects a request that is legal at the HTTP level — RFC 3986
+percent-encoding may carry arbitrary octets, unlike RFC 8259 which requires JSON to be valid
+UTF-8. Rejecting anyway is a deliberate call: sanitizing with `strings.ToValidUTF8` would keep
+such a request forwardable, but it costs an allocation and a scan per request on a path an
+attacker controls, and it makes the canonical method name lossy for stats and cache keys.
 
 ## Design
 
@@ -114,6 +121,38 @@ characters `\ud800` and passes the check. That is correct — escaped text is va
 cannot make `WithLabelValues` panic and cannot crash the process. Only genuinely raw invalid
 bytes are rejected, which is exactly the set that matters.
 
+### 4. The `chain` path parameter — `internal/server/http_server/http_server.go`
+
+Added in review. `chain := c.Param("chain")` is used verbatim as the label of
+`wsConnectionsMetric` (`ws_server.go:58,61`) with no validation anywhere in between, so it is
+the one remaining client-controlled string that reaches `WithLabelValues`. Every other `chain`
+label comes from `chain.String()` on a resolved `chains.Chain` enum and is safe by construction.
+
+The check goes in `requestHandler` **before `upgrader.Upgrade`**, not inside `HandleWebsocket`.
+That placement matters: once the connection is hijacked, `net/http`'s `conn.serve` recover
+skips its own cleanup (`if !c.hijacked()`), and `closeConn()` at `ws_server.go:158` is a plain
+call rather than a deferred one — so a panic after the upgrade leaks the socket and its FD with
+no read deadline, invisible to the gauge. Rejecting before the upgrade means no hijack, an
+ordinary parse-error response, and nothing to clean up. It also covers the plain HTTP path,
+which changes an invalid-UTF-8 chain's response from "chain … is not supported" to a parse
+error.
+
+An earlier draft of this document claimed the panic produced a clean connection drop with no
+leak. That was wrong, for the hijack reason above.
+
+#### Reaching it requires an uppercase escape
+
+Only `%FF` is a repro, not `%ff`. Go's escaper emits uppercase hex, so `%FF` round-trips: the
+re-escaped path equals the original, `url.Parse` leaves `URL.RawPath` empty, echo routes on the
+decoded `URL.Path`, and the param carries the raw `0xFF` byte. With `%ff` the re-escaped form
+differs, so `RawPath` is populated, echo routes on the encoded form, and the param is the three
+literal ASCII characters `%ff` — valid UTF-8, no panic. Measured:
+
+| request | `chain` param |
+| --- | --- |
+| `/queries/%FFeth/foo/%FFbar` | `"\xffeth"` |
+| `/queries/%ffeth/foo/%ffbar` | `"%ffeth"` |
+
 ## Cost
 
 Measured with `unicode/utf8` on Go 1.26 (Apple Silicon, 14 cores):
@@ -173,10 +212,12 @@ path, so the check cannot be used as an amplifier. No length cap or fast-path gu
   `NativeSubscribe`'s params, however, arrive as a `bytes` field with no such validation, so
   that path depends entirely on the new `getSubscription` check. Recorded here so this doesn't
   later get "discovered" as a hole.
-- **`ws_server.go:58,61` chain label.** Uses the raw `:chain` path parameter as a metric label —
-  a client-reachable invalid-UTF-8 label value this change deliberately does not fix, since it
-  is not a method name and so is outside this change's rule. Unlike the method sites, it runs
-  on the net/http handler goroutine, whose `conn.serve` recover catches the panic: the
-  connection drops with an "http: panic serving" log, the process survives, and no gauge leaks
-  (the deferred `Dec` never registers). It also requires auth to pass first. This is a known
-  follow-up, not a defect of this change.
+- **Unbounded `method` label cardinality.** `execution_flow.go:262` labels with
+  `request.Method()` before the `SpecMethod() == nil` check, so every distinct *valid-UTF-8*
+  unknown method mints a permanent series — and `GetSpecMethodWithFallback`
+  (`pkg/methods/helpers.go:52-58`) never returns nil for client JSON-RPC traffic, so that nil
+  check would not gate it even if the metric came after it. Same entry point and auth level as
+  the crash, unbounded memory rather than a restart. Raised in review; a real follow-up, and
+  bounding it needs more than reordering that one line.
+- **No `recover()` in the detached goroutine** at `execution_flow.go:259`. This change removes
+  the known trigger, but any future panic there still kills the process.
--- a/docs/superpowers/specs/2026-07-30-reject-non-utf8-methods-design.md
+++ b/docs/superpowers/specs/2026-07-30-reject-non-utf8-methods-design.md
@@ -1,0 +1,182 @@
+# Reject non-UTF-8 method names
+
+Date: 2026-07-30
+Branch: `reject_nonutf-8`
+
+## Problem
+
+RPC method names reach Prometheus label values and log lines. A client can put arbitrary
+bytes there:
+
+- JSON-RPC: `"method"` is a client-supplied JSON string. Neither `sonic` nor `encoding/json`
+  rejects raw invalid UTF-8 bytes inside a string literal.
+- REST: the method name falls back to `"<VERB>#/<restPath>"`, and `restPath` carries whatever
+  the client percent-encoded, after echo has decoded it.
+- WebSocket subscriptions: the `subscription` label of `nodecore_request_json_ws_connections`
+  comes from `getSubscription`, which for `eth_subscribe` reads `params[0]` verbatim.
+
+prometheus/client_golang validates label values unconditionally (`validateLabelValues`,
+`labels.go:177-181`), and `WithLabelValues` panics on invalid UTF-8 (`counter.go:283-289`,
+`gauge.go:237-243`) rather than letting a bad value into the registry — so the exposition
+itself is never corrupted. What actually happens is worse: `execution_flow.go:262` calls
+`requestTotalMetric.WithLabelValues(e.chain.String(), request.Method()).Inc()` as the first
+statement of a detached `go func()` in `processRequest`, and there is no `recover()` anywhere
+in `internal/`, `pkg/`, or `cmd/`. A panic there crashes the whole nodecore process. The same
+class of crash applies to the subscription label at `internal/upstreams/ws/registry_commands.go:79`.
+This change therefore fixes a remote-triggerable process crash — one HTTP request with a raw
+invalid byte in `"method"` — not a metrics-hygiene issue.
+
+Two pre-existing defects live in the same expression that has to be touched
+(`internal/upstreams/ws/ws_protocol.go:109` `getSubscription`):
+
+- `{"method":"eth_subscribe","params":[1]}` — `Raw()` returns `"1"`, so `sub[1:len(sub)-1]`
+  is `sub[1:0]` and **panics** with `slice bounds out of range [1:0]`. Verified locally.
+- `{"method":"eth_subscribe","params":[{"a":1}]}` — the label silently becomes `"a":1`.
+
+## Rule
+
+Reject a request if and only if its **method name** is not valid UTF-8. Nothing else is a
+reason to reject: invalid bytes in params, headers, query values, or REST wildcard captures
+pass through untouched, because they never become a method name or a metric label.
+
+## Design
+
+### 1. JSON-RPC — `internal/server/http_server/handlers.go`
+
+In `NewJsonRpcHandler`, after the sonic unmarshal succeeds, walk `jsonRpcRequests` and return
+a package-level sentinel error when any `Method` fails `utf8.ValidString`.
+
+The check lives in the constructor, not in `RequestDecode`, so HTTP and WS are both covered in
+one place and the request is refused before any work is scheduled.
+
+Consequences, both deliberate:
+
+- **Batches fail whole.** One bad method rejects the entire array with `protocol.ParseError()`
+  (`http_server.go:198-205`). Per-entry rejection would need response-holder machinery inside
+  `RequestDecode`; whole-batch rejection is how every other parse failure already behaves.
+- **On WebSocket the connection closes.** `ws_server.go:100-104` logs a constructor error and
+  breaks the read loop. This is the existing behavior for any parse failure and is kept as-is.
+
+### 2. REST — `internal/server/http_server/rest_parser.go`
+
+One check on `methodTemplate` immediately before `parseRestRequest` returns, covering both
+branches of the switch. `rest_parser.go` and `handlers.go` share the `http_server` package, so
+both sites return the same sentinel error.
+
+The matched-spec-template branch passes trivially — templates come from the embedded spec JSON.
+Only the `"<VERB>#/<restPath>"` fallback can reject. Junk bytes in a wildcard capture, a header,
+or a query value do not reject the request; they flow into `RequestParams` and the upstream URL
+as they do today.
+
+### 3. WS subscription type — `internal/upstreams/ws/ws_protocol.go`
+
+`getSubscription` gains an `error` return; `RequestFrame` (line 53) propagates it in the same
+`fmt.Errorf("... cause - %s")` style as its neighbours.
+
+```go
+if request.Method() == "eth_subscribe" {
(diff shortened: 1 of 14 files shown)


internal/upstreams/ws/ws_protocol.go
```diff
@@ -50,7 +52,12 @@ func (j *JsonRpcWsProtocol) RequestFrame(request protocol.RequestHolder) (*Reque
 
 	rawBody, _ := jsonBody.Raw()
 
-	return NewRequestFrame(requestId, getSubscription(&jsonBody, request), []byte(rawBody)), nil
+	subType, err := getSubscription(&jsonBody, request)
+	if err != nil {
+		return nil, fmt.Errorf("couldn't get a subscription type, cause - %s", err.Error())
```
tonatoz: `%s` + `err.Error()` severs the error chain, so `errors.Is(err, errNonUtf8SubType)` is false. That is why the new test can only assert `require.Error` and not the sentinel.

`internal/` uses `%w` in 88 places; `internal/upstreams/ws` uses it in zero — you are matching the file's local style, not the repo convention. Switch to `%w` here and tighten `TestJsonRpcWsProtocolRequestFrameRejectsNonUtf8SubType` to `require.ErrorIs`.

internal/server/http_server/rest_parser.go
```diff
@@ -50,6 +51,15 @@ func parseRestRequest(req *http.Request, restPath, specName string) (
...
 
+	// One check covers both switch branches. The matched-template branch passes
+	// trivially - templates come from the embedded spec JSON - so in practice this
+	// only ever rejects the "<VERB>#/<restPath>" fallback, where the client's bytes
+	// become the method name. Wildcard captures, headers, and query values are not
+	// checked: they never become a method name or a metric label.
+	if !utf8.ValidString(methodTemplate) {
+		return "", nil, errNonUtf8Method
+	}
+
```
tonatoz, line 62: This is over-rejection by the PR's own stated criterion.

For JSON-RPC, rejecting is semantically correct — RFC 8259 requires valid UTF-8, so the request was always malformed. For a URL path it is not: RFC 3986 percent-encoding carries arbitrary octets, and `GET /queries/<spec-without-rest-routes>/%FFfoo` is a legal request whose upstream URL is built from the literal path either way. Here a legal request is refused to protect a metric label.

Suggest sanitizing instead of rejecting on this branch:

```go
methodTemplate = strings.ToValidUTF8(fullPath, "\uFFFD")
```

That keeps the request forwardable, keeps the label safe, and keeps the canonical method name deterministic for stats/caching. No regression either way (pre-PR this crashed), but "over-rejection would be as much a defect as under-rejection" is your rule, and this is the one site where it bites.

Verified the branch is genuinely reachable — `/queries/%FFeth/foo/%FFbar` yields `chain="\xffeth"`, `rest="foo/\xffbar"`.

internal/server/http_server/handlers.go
```diff
@@ -133,6 +140,12 @@ func NewJsonRpcHandler(preReq *Request, requestBody io.Reader, isWsCtx bool) (*J
 		return nil, decoder.SyntaxError{}
 	}
 
+	for _, jsonRpcReq := range jsonRpcRequests {
+		if !utf8.ValidString(jsonRpcReq.Method) {
+			return nil, errNonUtf8Method
+		}
+	}
```
tonatoz: `for _, jsonRpcReq := range` copies a 64-byte `JsonRpcRequestBody` per entry to read one field. Index instead:

```go
for i := range jsonRpcRequests {
	if !utf8.ValidString(jsonRpcRequests[i].Method) {
		return nil, errNonUtf8Method
	}
}
```

Mutation-tested this guard: stubbing it out fails exactly `TestJsonRpcHandlerRejectsNonUtf8Method` and `...InBatch` and nothing else, so the coverage is real.

internal/server/http_server/handlers_test.go
```diff
@@ -144,3 +144,65 @@ func TestRestHandlerPromotesQueryAndHeadersIntoRequestParams(t *testing.T) {
...
+func TestJsonRpcHandlerRejectsNonUtf8Method(t *testing.T) {
+	body := "{\"id\":1,\"jsonrpc\":\"2.0\",\"method\":\"eth_\xffblockNumber\",\"params\":[]}"
+
+	handler, err := http_server.NewJsonRpcHandler(
+		&http_server.Request{Chain: "ethereum"},
+		strings.NewReader(body),
+		false,
+	)
+
+	require.EqualError(t, err, "method name is not a valid utf-8 string")
```
tonatoz: `require.EqualError` on literal text is brittle — an error-message reword breaks the test without signalling which invariant failed. It is also a pattern that appears nowhere else in this file.

Forced here only because the file is the external `http_server_test` package and cannot reach the unexported sentinel; `rest_parser_test.go` gets the clean `require.ErrorIs` precisely because it is an internal test file. Moving these two tests (this one and the batch one at :176) into an internal test file would make all three UTF-8 rejection tests consistent and drop the brittleness.

internal/server/http_server/rest_parser_test.go
```diff
@@ -219,3 +219,46 @@ func TestFilteredQuery_IsolatesSourceMutation(t *testing.T) {
...
+
+	require.ErrorIs(t, err, errNonUtf8Method)
+	assert.Empty(t, template)
+	assert.Nil(t, rp)
+}
+
+// A matched spec template is spec-provided and therefore always valid UTF-8. The
+// junk lives only in PathParams, which never becomes a method name - so the
+// request must NOT be rejected.
+func TestParseRestRequestAcceptsNonUtf8WildcardCapture(t *testing.T) {
```
tonatoz: Naming: the three new tests drop the file's `TestParseRestRequest_<Scenario>` underscore convention that all ~14 existing tests follow (`_MatchedLiteralTemplate`, `_WildcardCapturesPathParam`, `_UnmatchedFallsBackToLiteral`, ...). Suggest `TestParseRestRequest_RejectsNonUtf8MethodTemplate` etc.

docs/superpowers/specs/2026-07-30-reject-non-utf8-methods-design.md
```diff
@@ -0,0 +1,182 @@
...
+prometheus/client_golang validates label values unconditionally (`validateLabelValues`,
+`labels.go:177-181`), and `WithLabelValues` panics on invalid UTF-8 (`counter.go:283-289`,
+`gauge.go:237-243`) rather than letting a bad value into the registry — so the exposition
+itself is never corrupted. What actually happens is worse: `execution_flow.go:262` calls
+`requestTotalMetric.WithLabelValues(e.chain.String(), request.Method()).Inc()` as the first
+statement of a detached `go func()` in `processRequest`, and there is no `recover()` anywhere
+in `internal/`, `pkg/`, or `cmd/`. A panic there crashes the whole nodecore process. The same
+class of crash applies to the subscription label at `internal/upstreams/ws/registry_commands.go:79`.
+This change therefore fixes a remote-triggerable process crash — one HTTP request with a raw
+invalid byte in `"method"` — not a metrics-hygiene issue.
```
tonatoz, line 27: Agreed on the diagnosis, and I verified it end to end: sonic decodes `{"method":"eth_\xffblockNumber"}` with `err == nil` and `utf8.ValidString == false`, and `WithLabelValues` then panics with `label value "eth_\xffblockNumber" is not valid UTF-8`.

But I'd argue the deferred `recover()` is the load-bearing fix and this PR is the mitigation, so it should not wait:

1. **Same line, worse bug, not listed in Out of scope.** The metric fires *before* the `SpecMethod() == nil` check (262 vs 264), so every distinct **valid-UTF-8** unknown method mints a permanent series in `requestTotalMetric` — and again in `requestErrorsMetric`, `dims`, `caches`, `ratelimiter`, `rating`. Same auth level, same entry point, unbounded memory plus scrape blowup. Out of scope covers *subscription* label cardinality but not the method label right next to it. A crash restarts; this degrades silently.
2. Moving line 262 below the nil check is a one-liner, and `GetMetricWithLabelValues` (returns an error instead of panicking) makes the whole class non-fatal.
3. Validating at N parse sites needs an N+1th site for every future label. The `chain` gap below is already that N+1.

docs/superpowers/specs/2026-07-30-reject-non-utf8-methods-design.md
```diff
@@ -0,0 +1,182 @@
...
+  `NativeSubscribe`'s params, however, arrive as a `bytes` field with no such validation, so
+  that path depends entirely on the new `getSubscription` check. Recorded here so this doesn't
+  later get "discovered" as a hole.
+- **`ws_server.go:58,61` chain label.** Uses the raw `:chain` path parameter as a metric label —
+  a client-reachable invalid-UTF-8 label value this change deliberately does not fix, since it
+  is not a method name and so is outside this change's rule. Unlike the method sites, it runs
+  on the net/http handler goroutine, whose `conn.serve` recover catches the panic: the
+  connection drops with an "http: panic serving" log, the process survives, and no gauge leaks
+  (the deferred `Dec` never registers). It also requires auth to pass first. This is a known
+  follow-up, not a defect of this change.
```
tonatoz: "The connection drops, the process survives, and no gauge leaks" is only half right — and I'd correct it here rather than let it stand as the follow-up's premise.

`HandleWebsocket` runs synchronously on the echo handler goroutine, so `net/http`'s recover does catch it (`server.go:1903`). But the cleanup guard is `if !c.hijacked()` (`server.go:1914`), and `upgrader.Upgrade` already hijacked the connection — so `c.close()` is skipped. And `closeConn()` at `ws_server.go:158` is a **plain call, not deferred**, so it never runs either. The socket and its FD leak, with no read deadline, invisible to the gauge.

Not process death, but not a clean drop. `defer closeConn()` is the one-line fix and belongs with the recover work above. The gauge-leak reasoning is right: the panic at `:58` precedes the `defer` at `:60`.

docs/superpowers/specs/2026-07-30-reject-non-utf8-methods-design.md
```diff
@@ -0,0 +1,182 @@
...
+
+## Problem
+
+RPC method names reach Prometheus label values and log lines. A client can put arbitrary
+bytes there:
+
+- JSON-RPC: `"method"` is a client-supplied JSON string. Neither `sonic` nor `encoding/json`
+  rejects raw invalid UTF-8 bytes inside a string literal.
+- REST: the method name falls back to `"<VERB>#/<restPath>"`, and `restPath` carries whatever
+  the client percent-encoded, after echo has decoded it.
```
tonatoz, line 14: Worth stating that this only holds for **uppercase** percent-escapes. With `%FF`, Go leaves `URL.RawPath` empty, echo routes on `URL.Path`, and the param carries the raw `0xFF` byte. With `%ff`, `RawPath` is populated, echo routes on the encoded form, and the param is the literal three ASCII characters `%ff` — valid UTF-8, no panic.

Measured: `/queries/%FFeth/foo/%FFbar` -> `chain="\xffeth"`, `rest="foo/\xffbar"`; `/queries/%ffeth/foo/%ffbar` -> `chain="%ffeth"`, `rest="foo/%ffbar"`. The PR body's `%FFfoo` example is right, but the reader should know the lowercase form is not a repro.

docs/superpowers/specs/2026-07-30-reject-non-utf8-methods-design.md
```diff
@@ -0,0 +1,182 @@
...
+| CJK method name (63 B, multi-byte) | 22.7 ns |
+| 1 KB ASCII | 14.9 ns |
+| 1 MB ASCII (pathological method name) | 11 µs |
+| 1 MB with invalid first byte | 1.7 ns (early exit) |
+
+No allocations — the strings already exist and `ValidString` does not copy. The pathological
+case is bounded by work already done: sonic had to parse and allocate that same payload to hand
+it over. Across a batch, the sum of method lengths is at most the body size, so validation stays
+O(body) — a small constant factor on top of the parse that just ran. Rejection is the cheapest
+path, so the check cannot be used as an amplifier. No length cap or fast-path guard is needed.
```
tonatoz, line 173: This section benchmarks a check nobody would question. The interesting number is missing: series count and registry memory under a flood of distinct *valid* method names — which is the actual remaining exposure at `execution_flow.go:262`.

tonatoz (changes requested):

internal/server/http_server/rest_parser.go
KirillPamPam: For now I suggest to check the path and also reject such requests. In that case it's easy to ddos a service with non valid paths and it takes a while to perform ToValidUTF8 then. For simplicity let's just reject them.

docs/superpowers/specs/2026-07-30-reject-non-utf8-methods-design.md
KirillPamPam: Moving a line doesn't solve a problem because basically there is always SpecMethod, for unknown ones it calls a fallback default SpecMethod

internal/upstreams/ws/ws_protocol.go
KirillPamPam: Fixed

internal/server/http_server/handlers.go
KirillPamPam: Agree, thanks!

internal/server/http_server/handlers_test.go
KirillPamPam: Fixed

internal/server/http_server/rest_parser_test.go
KirillPamPam: Done

docs/superpowers/specs/2026-07-30-reject-non-utf8-methods-design.md
KirillPamPam: Added a check above - now if chain is non valid utf8, then fail

docs/superpowers/specs/2026-07-30-reject-non-utf8-methods-design.md
KirillPamPam: Added

tonatoz (approved):

PR #339, opened 2026-08-20
Sui grpc support

KirillPamPam:
# gRPC support v1: Sui (unary-only)

## Summary

Adds **native gRPC as a first-class upstream protocol**, routed through the same execution flow as
JSON-RPC and REST, and onboards **Sui** (`sui.rpc.v2`, mainnet + testnet) as the first gRPC-native
chain. v1 is a self-contained vertical slice:

```
gRPC client (real Sui stubs / grpcurl / Postman) → nodecore chain ingress → execution flow → GrpcConnector → Sui node
```

The gRPC connector is the Sui upstream's **only** connector and owns heads, health, chain
validation, labels and lower bounds. The ingress runs as a **separate gRPC server** on a new
`server.grpc-ingress-port`, serves **full gRPC reflection** for every routable chain method (grpcurl
and Postman work against nodecore exactly as against a native node), and shares the request pipeline
(auth → decode → chain checks → flow) with the HTTP/WS servers via one extracted function.

**Zero emerald-grpc changes** — NativeCall/NativeSubscribe and the dshackle server are untouched;
the dproxy ingress is a designed follow-up. Design doc: `docs/superpowers/specs/2026-08-19-grpc-support-v1-design.md`.

## Core principles

- **The traffic path is bytes-only.** A gRPC call is a method string
  (`/sui.rpc.v2.LedgerService/GetObject`), serialized protobuf treated as a black box, metadata
  (= HTTP/2 headers), and a closed 17-code status model. That is exactly nodecore's
  `Method()` + `Body()` abstraction — no schemas in the connector/flow/protocol layers.
- **Arity is not on the gRPC wire.** Unary vs server-stream is not encoded in the protocol — the
  method specs are the only source of it.
- **Chain and auth ride metadata** (`x-nodecore-chain`, `x-nodecore-key`): the chain cannot go in
  `:path` (gRPC owns it), and service names cannot identify chains (every Cosmos-SDK chain shares
  them).
- **Schema boundary.** Generated proto models (`pkg/sui`) exist for the chain-specific probes; they
  cross the boundary as bytes (`proto.Marshal` into a request body, `proto.Unmarshal` of response
  bytes).

### Method specs (`pkg/methods`)

New extension point: `MethodSettings.Grpc *GrpcSettings{CallType}` with `call-type` ∈
{`unary`, `server-stream`}; absent block = unary, only streaming methods carry the annotation.
Validation: the enum, `grpc` block only in specs whose `api-connectors` include `grpc`, method names
must match the `/package.Service/Method` shape, and `server-stream` is mutually exclusive with
`sticky`/`dispatch`/`cacheable: true` (server-stream methods also default to non-cacheable).

`sui-grpc.json` lists all **27 methods of the 7 `sui.rpc.v2` services** (verified against
MystenLabs/sui-apis protos): 21 unary + 6 server-stream (`ListCheckpoints/Transactions/Events`,
`Subscribe*`). Streaming methods are deliberately **listed** so the ingress answers
"server-streaming not supported yet" instead of "unknown method"; client-streaming/bidi are never
listed — absence is the rejection. Everything is `cacheable: false` (gRPC caching is a follow-up).
`GetGrpcServices()` exposes the spec-declared service names — the reflection advertisement source.

### Protocol plumbing (`internal/protocol`)

- `UpstreamGrpcRequest`: method string verbatim, body = the serialized request message exactly as
  the client produced it (the 5-byte gRPC wire frame lives below the codec and never appears here),
  metadata in `RequestParams.Headers`, request hash over method + body + selector label key with
  `calculateRestHash`'s injective framing.
- **Two-track status mapping** — classification drives routing, the original status (code +
  message + typed details as a serialized `google.rpc.Status`) rides through verbatim in a typed
  `GrpcStatus` on `ResponseError.Data`:
  - client, no retry: CANCELLED, INVALID_ARGUMENT, NOT_FOUND, ALREADY_EXISTS, FAILED_PRECONDITION, OUT_OF_RANGE
  - retry on another upstream: UNKNOWN, INTERNAL, DATA_LOSS, UNAVAILABLE, ABORTED
  - retry/hedge eligible: DEADLINE_EXCEEDED; throttled (feeds auto-tune): RESOURCE_EXHAUSTED
  - ban method on the upstream + retry elsewhere: UNIMPLEMENTED (`ClassifyMethodAvailability` is
    gRPC-aware; JSON-RPC message regexes never run against gRPC statuses)
  - upstream auth/config problem, pass through: PERMISSION_DENIED, UNAUTHENTICATED
- `ResponseError.Code` for gRPC errors is offset by `GrpcErrorCodeBase = 10000`: canonical codes
  1–16 would otherwise alias nodecore's own error constants (e.g. CANCELLED=1 vs
  `NoAvailableUpstreams`, DEADLINE_EXCEEDED=4 vs `WsTotalFailure`). The int is informational; all
  routing and client-facing rendering read the typed `GrpcStatus`.
- `HasResponseTrailers` — a sibling of `HasResponseHeaders`: a gRPC client must receive trailers
  *as* trailers.

### GrpcConnector (`internal/upstreams/connectors`)

Config shape unchanged (`type: grpc`, `url`, `headers` → per-call metadata, `ca`). URL scheme picks
transport security: `grpc://`/`http://` = plaintext, anything else = TLS (system roots or `ca`).
One long-lived `ClientConn` per connector (HTTP/2 multiplexing, client keepalives 30s/10s).

`SendRequest` = `conn.Invoke` with a **raw passthrough codec whose `Name()` is `"proto"`** — the
content-type stays `application/grpc+proto`, so the node's stock stubs decode our bytes as the
normal protobuf they are. Client metadata is forwarded minus a deny list (reserved `grpc-*` family +
hop-by-hop + config-owned keys); response **headers and trailers** are captured, filtered
(HTTP deny defaults + `content-type` + `grpc-*`, extensible via `response-header-deny`) and carried
on the response holder. Errors arrive as `*status.Status` (they travel in trailers) regardless of
codec.

Client hygiene: **60s per-call cap** (mirrors the HTTP connector; deadlines only tighten, so probe
timeouts and client deadlines stay in force — and the cap propagates to the node as `grpc-timeout`),
with the caller's own deadline distinguished from the cap (caller gave up → total failure; our cap →
retryable DEADLINE_EXCEEDED); **32MB receive limit** (a proxy must not impose a lower response
ceiling than the node; grpc-go's default is 4MB); **reconnect backoff capped at 30s** (default
climbs to 120s — a recovered upstream should rejoin rating sooner); fail-fast (no wait-for-ready —
failures feed the flow's retry to another upstream); no transport-level retries (the flow owns
retry/hedge). `Subscribe` = not supported, `SubscribeStates` = nil (connectivity-state mapping is a
follow-up).

### Sui chain family (`pkg/sui`, `internal/upstreams/.../sui_*`)

- Registry: `chains.Sui` type; `SUI` / `SUI_TESTNET` regenerated from chains.yaml (grpcIds
  1176/10210, chain-ids = genesis checkpoint digests).
- **Models pipeline**: `chain-apis/sui` submodule (MystenLabs/sui-apis pinned to `201981c` — the
  repo has no release tags) + `chain-apis/sui.gen.yaml` (buf v2, managed `go_package`) + one generic
  Makefile pattern rule (`make sui-proto-gen`; a future chain adds its submodule + yaml, zero
  Makefile edits). `protoc-gen-go` is a go.mod **tool directive**, so generator and runtime can
  never drift. The **whole `sui/rpc/v2` subtree is generated** (33 files, message types only, no
  client stubs): the probes only need the `ledger_service.proto` closure to compile, but ingress
  reflection serves the descriptors these files register, and reflection must cover every routable
  service — codegen is per-file, and dynamic/hand-written alternatives were evaluated and rejected
  in the design doc. `.gitattributes` marks all generated files `linguist-generated` (collapsed in
  review).
- `sui_specific`: every probe is a unary gRPC call reading the **single source**
  `LedgerService/GetServiceInfo` (empty request — zero bytes on the wire). Head =
  `checkpoint_height` with **synthetic height-derived hashes** (the Stellar/Solana pattern —
  GetServiceInfo exposes no digest; checkpoints are BFT-final, no reorgs, so height-derived ids are
  safe and parent-linkable); finalized = latest (an executed checkpoint *is* final); health = one
  validator (transport/status error → Unavailable); chain validation compares `chain_id` (genesis
  digest, case-insensitive — the registry lowercases chain-ids) — stricter than the network-name
  compare; labels parse only `server` ("sui-node/1.78.0-commit" → client `sui-node`, version
  `1.78.0`, commit trimmed like Stellar); lower bounds map `lowest_available_checkpoint` →
  BlockBound and `..._objects` → StateBound. Poll-only head (no subscription in v1), no cap
  detectors, no methods introspection.

### Chain ingress (`internal/server/grpc_ingress`, `grpc_server`, `server_ctx`)

A **separate gRPC server** on `server.grpc-ingress-port` (0/absent = disabled), reusing `server.tls`;
`internal/server/grpc_server` is now a thin lifecycle package managing both endpoints (dshackle on
`grpc-port`, ingress on `grpc-ingress-port`; either runs alone). Separate on purpose: different
clients, different auth (client keys/tokens vs dshackle sessions), different codec needs, keepalive
enforcement scoped to ingress connections only, and each port advertises its own coherent
reflection surface. This was the original design's own named fallback, promoted — it also voids the
delegating-codec benchmark obligation, since the codec no longer touches dshackle traffic.

- **Delegating codec**: a `grpc.Server` has exactly one server-wide codec, and this server hosts
  two worlds — the reflection service (generated code, needs the proto codec) and the catch-all
  (needs raw bytes). The codec dispatches on the value type the handler passed (`*rawFrame` → bytes
  as-is, else the wrapped default proto `CodecV2`, buffer pooling preserved).
- **Catch-all** (`UnknownServiceHandler`): `emerald.*` guard → UNIMPLEMENTED; `Authenticate` over a
  new `GrpcAuthPayload` (key, token and **JWT** strategies all work natively over metadata:
  `x-nodecore-key`, `x-nodecore-token`, `authorization: Bearer …`); then the shared pipeline. The
  decode reads the chain from metadata, resolves arity from the spec (unknown method and
  server-stream → UNIMPLEMENTED), and reads **exactly one request message, dispatching immediately —
  never waiting for the client's half-close** (grpc-go's own `processUnaryRPC` does the same).
  Response: upstream headers via `SetHeader`, trailers via `SetTrailer`, one `SendMsg`; errors as
  the upstream's **verbatim `*status.Status`, typed details included**.
- **Shared pipeline**: `http_server.handleRequest` moved verbatim to
  `server_ctx.HandleRequest` (PreKeyValidate → decode → chain checks → PostKeyValidate → flow with
  the canonical hook set) behind a minimal two-method `RequestHandler`; HTTP/WS and gRPC are now the
  same pipeline with per-protocol decode/encode. The hook list exists in exactly one place.
- **Reflection**: registered (v1 + v1alpha) with a `ServiceInfoProvider` that advertises the
  spec-declared chain services on top of the registered ones — names come from the specs, content
  from the descriptors linked into the binary (`chain_descriptors.go` is the explicit blank-import
  anchor; an invariant test fails CI with the exact missing package if a spec and the anchor ever
  diverge). `ListServices`, `describe` and invoke work in grpcurl and Postman against the full Sui
  surface. Module-backed chains (e.g. Cosmos via `cosmossdk.io/api`) onboard descriptors with
  blank imports only — no vendoring, no codegen.
- Server keepalives (2m/20s) + enforcement (`MinTime: 10s`) on the ingress server only.

## Config

```yaml
server:
  grpc-ingress-port: 9095   # separate from grpc-port (dshackle); reuses server.tls

upstreams:
  - id: sui-1
    chain: sui
    connectors:
      - type: grpc
        url: grpcs://sui-mainnet.example.com:443   # grpc:// = plaintext, grpcs://|https:// = TLS
```

```bash
grpcurl -H 'x-nodecore-chain: sui' localhost:9095 list
grpcurl -H 'x-nodecore-chain: sui' -d '{}' localhost:9095 sui.rpc.v2.LedgerService/GetServiceInfo
```

## Testing

- Unit tests per task; connector and ingress tested end-to-end against in-process grpc-go servers
  over `bufconn` with **real generated stubs on the peer side** — typed client → catch-all → mocked
  upstream, verbatim status/details round-trips, metadata deny lists, header/trailer buckets,
  reflection (ListServices + FileContainingSymbol), key/token auth over metadata, cap-vs-caller
  deadline classification, >4MB responses.
- `sui_specific` probe tests run on `pkg/sui`-marshaled fixtures.
- The HTTP/WS suite passing untouched is the proof the pipeline extraction was mechanical.
- `make test` (full `-race` suite) and `golangci-lint` green.


internal/upstreams/connectors/grpc_connector.go
```diff
@@ -0,0 +1,339 @@
...
+	"te",
+	"trailer",
+	"trailers",
+	"transfer-encoding",
+	"upgrade",
+	"host",
+	"content-length",
+	"content-type",
+	"accept-encoding",
+	"user-agent",
```
tonatoz, line 62: The request-side deny list covers hop-by-hop keys and the reserved `grpc-*` family, but not nodecore's own metadata, so client credentials are forwarded to the node. Verified against the bufconn harness in this package (peer server dumping its incoming metadata):

```
upstream saw x-nodecore-key=[client-api-key] token=[client-secret-token]
             authorization=[Bearer client-jwt] chain=[sui]
```

For third-party upstreams that leaks the end client's api key / static token / JWT, plus `x-nodecore-chain`, which is meaningless to the node. Suggest adding `x-nodecore-key`, `x-nodecore-token`, `x-nodecore-chain` and `authorization` to `grpcRequestMetadataDeny` (config-owned headers already win over client-supplied ones, so an upstream that needs its own `authorization` keeps working).

Side note: `defaultRequestHeaderDeny` in the HTTP connector has the same gap on the REST path (where `RequestParams.Headers` is populated), so it would be consistent to close both.

internal/upstreams/connectors/grpc_connector.go
```diff
@@ -0,0 +1,339 @@
...
+			if statusProto, marshalErr := proto.Marshal(st.Proto()); marshalErr == nil {
+				grpcStatus.StatusProto = statusProto
+			}
+		}
+		response := protocol.NewGrpcUpstreamErrorResponse(request, grpcStatus)
+		if generic, isGeneric := response.(*protocol.GenericUpstreamResponse); isGeneric {
+			generic.WithResponseHeaders(g.filterResponseMetadata(headerMD)).
+				WithResponseTrailers(g.filterResponseMetadata(trailerMD))
+		}
+		return response
```
tonatoz, line 209: `NewGrpcUpstreamErrorResponse` returns `*ReplyError` for the partial-failure branch (UNKNOWN, INTERNAL, DATA_LOSS, UNAVAILABLE, ABORTED, DEADLINE_EXCEEDED, RESOURCE_EXHAUSTED, UNIMPLEMENTED), so this type assertion fails and header/trailer metadata is dropped for exactly those statuses — the ones most likely to carry `retry-after` / rate-limit hints that the client would want when the flow gives up after retries.

Also, the shape is fragile: the constructor returns an interface and the caller type-asserts to attach fields, so any future error-response type silently loses metadata. Consider hanging the header/trailer buckets on `ReplyError` too (or having `NewGrpcUpstreamErrorResponse` take them as arguments) so the behaviour is uniform and enforced by the type.

internal/server/grpc_ingress/chain_ingress.go
```diff
@@ -0,0 +1,212 @@
...
+		grpc.KeepaliveParams(keepalive.ServerParameters{
+			Time:              2 * time.Minute,
+			Timeout:           20 * time.Second,
+			MaxConnectionIdle: 3 * time.Minute,
+		}),
+		grpc.KeepaliveEnforcementPolicy(keepalive.EnforcementPolicy{
+			MinTime:             10 * time.Second,
+			PermitWithoutStream: true,
+		}),
+	)
```
tonatoz, line 61: This is a new externally reachable port and the server options set no stream ceiling: grpc-go's default is `maxConcurrentStreams: math.MaxUint32` (`server.go:189` in grpc v1.83), and the catch-all blocks in `RecvMsg` (line 138) with no server-side deadline of its own.

The keepalive settings do not bound that: a connection with an open stream is not idle, so `MaxConnectionIdle` never fires, and the keepalive ping only detects dead transports, not a client that opens streams and never sends a message. So one TCP connection can park an unbounded number of handler goroutines plus HTTP/2 stream state.

Suggest `grpc.MaxConcurrentStreams(N)` plus a short deadline around the first `RecvMsg` (or a `grpc.ConnectionTimeout`-style bound), so an idle-but-open stream cannot be held for free.

internal/server/grpc_ingress/chain_ingress.go
```diff
@@ -0,0 +1,212 @@
...
+	authPayload := auth.NewGrpcAuthPayload(md)
+	if err := c.appCtx.AuthProcessor.Authenticate(ctx, authPayload); err != nil {
+		return status.Errorf(codes.Unauthenticated, "auth error - %s", err.Error())
+	}
+
+	requestHandler := &grpcRequestHandler{stream: stream, md: md, method: fullMethod}
+	handleResp := c.appCtx.HandleRequest(ctx, requestHandler, authPayload, flow.NewSubCtx())
+	wrapper, ok := <-handleResp.ResponseWrappers()
+	if !ok {
+		return status.FromContextError(ctx.Err()).Err()
```
tonatoz: `status.FromContextError(nil)` returns a nil `*Status`, and `(*Status)(nil).Err()` returns nil (nil status reports `codes.OK`). So if the response channel is closed without a wrapper while `ctx.Err() == nil`, the handler returns nil, the server sends status OK with no message, and the client's unary call fails with a bare `io.EOF` instead of a status.

Worth an explicit fallback, e.g. keep the context-error mapping when `ctx.Err() != nil` and otherwise return `status.Error(codes.Internal, "no response from the execution flow")`.

internal/upstreams/upstream_factory.go
```diff
@@ -154,6 +155,8 @@ func createConnector(
 		return connectors.NewHttpConnector(connectorConfig, specs.RestIndexer, torProxyUrl, upId)
 	case specs.RestAdditional:
 		return connectors.NewHttpConnector(connectorConfig, specs.RestAdditional, torProxyUrl, upId)
+	case specs.GrpcConnector:
+		return connectors.NewGrpcConnector(connectorConfig, upId)
```
tonatoz, line 159: `torProxyUrl` is in scope here but not passed to `NewGrpcConnector`, and the gRPC connector dials directly. Meanwhile `ApiConnectorConfig.validate` accepts a `.onion` host as long as `server.tor-url` is set, so `type: grpc` + an onion URL passes config validation and then fails at dial time with no hint about the cause.

Either reject the combination in validation (grpc connector + onion host) or wire a proxy dialer into `grpc.NewClient` via `grpc.WithContextDialer`.

internal/upstreams/connectors/grpc_connector.go
KirillPamPam: Fixed

internal/upstreams/connectors/grpc_connector.go
KirillPamPam: Fixed

internal/server/grpc_ingress/chain_ingress.go
KirillPamPam: Added a timeout for `RecvMsg`

internal/server/grpc_ingress/chain_ingress.go
KirillPamPam: fixed

internal/upstreams/upstream_factory.go
KirillPamPam: Reject the combination in validation

tonatoz (approved):

PR #346, opened 2026-08-25
Support grpc server streaming

KirillPamPam:
# gRPC server streaming: `Subscribe*` / `List*` pass-through and a subscription-driven head for gRPC upstreams

## Summary

Server-streaming gRPC calls now flow end to end — client → gRPC chain ingress → execution flow →
`GrpcConnector` → node, and back — bytes-only, one upstream stream per client. Two kinds of stream
are supported and told apart by the spec:

- **subscriptions** (`sui.rpc.v2.SubscriptionService/Subscribe*`): live, unbounded; the node ending
  the stream is a failure (`UNAVAILABLE`);
- **finite streams** (`LedgerService/List*`): a result delivered as several messages; the node
  ending the stream is normal completion (`OK`, trailers included).

A gRPC upstream can additionally track its head via a head subscription instead of polling
(`head-mode`), and Sui does so over `SubscribeCheckpoints`.

Follows #339 (unary gRPC) and #345 (transport-neutral `SubResponse`). Design:
`docs/superpowers/specs/2026-08-25-grpc-server-streaming-design.md`.

## Explicitly out of scope

- Aggregation of gRPC streams: every client gets its own upstream stream (see `resolveSource`, the
  branch is marked `TEMPORARY`).
- Client-streaming and bidi calls (not in any spec; absence is the rejection).
- Retries, hedges, caching, quorum for streams. Connector state events for gRPC
  (`SubscribeStates` stays nil).

## Method specs

`grpc.call-type` gains two values and loses `server-stream`:

| value | meaning | `Method.IsSubscribe()` |
|---|---|---|
| `unary` (default) | one request, one response | false |
| `server-stream-subscription` | unbounded live stream | true |
| `server-stream-finite` | bounded stream | true |

Both streaming kinds are subscriptions for the router — no failsafe, `SubscriptionRequestProcessor`,
streamed responses — so nothing in the flow needed a new branch. The only difference, the meaning of
a clean EOF, is decided in exactly one place (`newGenericSourceBuilder`).

`specs.IsSubscribeMethod` used to read the `Subscription` settings block directly (a second
definition of "is a subscription"); it is now a wrapper over `Method.IsSubscribe()`. `subscription`
settings are rejected on gRPC methods — `grpc.call-type` carries that information.
`UpstreamGrpcRequest.IsSubscribe()` follows the spec instead of returning `false`.

## Connector

`GrpcConnector.Subscribe` opens the stream (`NewStream` → one `SendMsg` → `CloseSend`, all
synchronous so open failures are returned from `Subscribe` as the verbatim gRPC status) and runs one
receive goroutine that forwards every frame as it arrives. The 60s unary cap does not apply; the
stream lives exactly as long as the ctx passed to `Subscribe`, so `Unsubscribe` is a no-op. Sends
block (never drop): HTTP/2 flow control pauses the node while the consumer is busy, and both
consumers — the subengine fan-out and the subscription head — drain continuously.

**End of stream is a frame.** `SubResponse` gains a third kind next to data and error: `IsEnd()`.
The connector is kind-agnostic: message → data frame, status → error frame, `io.EOF` → end frame.
Trailers ride on the terminal frame (error or end), headers on the first frame whichever kind it is.

## Flow

- `SubscriptionRequestProcessor` is reused for both kinds. The mode-dependent parts are now a
  strategy, `subFraming` (`sub_framing.go`): `jsonRpcFraming` (WS: subscription id, ack, notification
  envelope) and `resultOnlyFraming` (gRPC ingress, emerald server: bare payloads plus the frame's
  gRPC headers/trailers). `ProcessRequest` is linear: guard → resolve → subscribe → `begin` → loop →
  terminal.
- The guard is now `!SpecMethod().IsSubscribe()`: it refuses plain methods *and* unsubscribe calls
  flagged as subscriptions (the emerald server flags every native-subscribe request). The old
  `Subscription == nil` guard let unsubscribe methods through.
- `newGenericSourceBuilder` forwards an end frame for a finite stream and turns it into
  `SubscribeTotalFailure` for a subscription. gRPC streams get a per-request engine key
  (`RequestHash|selectors|uuid`) so the engine never shares a source — a late joiner of a
  `ListCheckpoints` would otherwise miss its first messages.
- `WsCapMatcher` applies to JSON-RPC subscriptions only; a gRPC stream rides the grpc connector the
  spec binds the method to.

## Subscription engine

- An end frame terminates a source like an error frame does. `Subscription.Err()` is derived from the
  terminal frame, and `Subscription.Terminal()` exposes the frame itself so transport metadata
  (trailers on a `RESOURCE_EXHAUSTED` frame, trailers on a clean end) survives the fan-out. The
  processor turns it into the client's final response: the `ReplyError` with metadata, or the new
  non-event `SubscriptionEndResponse`.
- `Source.Exclusive`: a never-shared source (per-request key) is torn down as soon as its subscriber
  leaves instead of after the 10s reuse grace, so a cancelled client releases the upstream stream
  immediately.
- A channel closed without a terminal frame still means total failure — the locally-synthesized
  sources rely on that and are unaffected. WS behaviour is unchanged throughout (WS sources never
  emit end frames; `Err()` values are identical).

## Ingress

`chainIngress.handle` picks a call shape via a private `grpcCall` seam — `unaryCall` (today's body)
and `serverStreamCall` (relay event frames, forward metadata, return the upstream's status or `OK`
on a clean close; `SubscribeTotalFailure → UNAVAILABLE`). Both share the one-frame request decode;
the seam is where client-streaming/bidi would land later. The "server-streaming methods are not
supported yet" rejection is gone.

## Head: `head-mode`

New upstream setting `head-mode: subscribe | poll` (default `subscribe`), consulted only when the
head connector is `grpc` (other connector types imply their mode). `createHead` calls
`SubscribeHeadRequest()` once: `blocks.ErrUnsupportedHeadSubscriptions` → warn and poll; any other
error → a subscription head that fails loudly on `Start`. No runtime fallback: a node without
`SubscriptionService` (Sui answers `UNIMPLEMENTED`) needs `head-mode: poll` — documented.

`blocks.ErrUnsupportedHeadSubscriptions` unifies what used to be eight private per-chain sentinels,
four ad-hoc `fmt.Errorf`s and tron REST's `nil, nil` (a latent bug: a subscription head over it
would have called `Subscribe(ctx, nil)`).

Sui: `SubscribeHeadRequest` opens an unfiltered `SubscribeCheckpoints` with `read_mask:
sequence_number`; `ParseSubscriptionBlock` takes the height from `cursor` (present on every frame,
monotonic), so the checkpoint payload is never inspected. Hashes stay synthetic, matching the polled
head.

## Client-visible behaviour

| situation | client sees |
|---|---|
| node rejects the stream at open | the node's status, details included |
| node aborts mid-stream | the node's status, headers/trailers forwarded |
| node ends a subscription cleanly | `UNAVAILABLE` |
| node ends a finite stream cleanly | `OK` with trailers |
| client too slow (engine fan-out buffer full) | `UNAVAILABLE` (`SubscriberTooSlow`) |
| client cancels | upstream stream cancelled at once, nothing sent |

## Docs

`05-upstream-config.md` (`head-mode`, gRPC connector paragraph), `11-method-specs.md` (call types),
`14-grpc-ingress.md` (streams, error table, limitations).


internal/upstreams/flow/sub_processor.go
```diff
@@ -111,16 +106,10 @@ func (s *SubscriptionRequestProcessor) ProcessRequest(
 				if filter != nil && !filter.Matches(r.GetParsedEvent()) {
 					continue
 				}
-				var subResponse protocol.ResponseHolder
-				if s.subCtx.IsSubscriptionResultOnly() {
-					subResponse = protocol.NewSubscriptionResultEventResponse(request.Id(), r.GetMessage())
-				} else {
-					subResponse = protocol.NewSubscriptionMethodResultResponse(request.Id(), method, r.GetMessage(), subId)
-				}
 				responses <- &protocol.ResponseHolderWrapper{
```
tonatoz: This send (and the ones at lines 81, 85 and 99) has no ctx escape, so it parks the goroutine forever once the client stops reading. `serverStreamCall.serve` returns as soon as `ctx.Done()` fires, `SendMsg` fails or an error frame arrives, so with a wrapper in flight nobody ever drains `responses` and the deferred `sub.Unsubscribe()` never runs.

Reproduced on this branch: push two frames, do not read the wrappers, cancel the ctx, dump the stacks:

```
goroutine ... flow.(*SubscriptionRequestProcessor).ProcessRequest.func1()
	internal/upstreams/flow/sub_processor.go:109   <- chan send
```

Because the engine actor's `srcCtx` derives from `e.ctx` (the per-chain registry ctx) and not from the request ctx, `refs` stays at 1, the exclusive source is never terminated and the upstream gRPC stream to the node stays open until the 100-slot subscriber buffer overflows into `SubscriberTooSlow` - i.e. never for a quiet or stalled stream. That makes the "client cancels -> upstream stream cancelled at once, nothing sent" row conditional: it holds only when the select happens to take the `execCtx.Done()` branch. `TestChainIngressClientCancelStopsTheUpstreamStream` passes because no frame is pending at cancel time.

Suggested shape for every send in this goroutine:

```go
select {
case responses <- wrapper:
case <-execCtx.Done():
    return
}
```

internal/upstreams/flow/sub_aggregation.go
```diff
@@ -268,8 +295,14 @@ func newGenericSourceBuilder(
 						out <- &protocol.GenericSubResponse{Error: protocol.SubscribeTotalFailureError(), UpstreamId: upstreamId}
 						return
 					}
+					if r.IsEnd() && !finite {
+						// a node ending a live subscription is a failure; only a
+						// bounded stream completes
+						out <- &protocol.GenericSubResponse{Error: protocol.SubscribeTotalFailureError(), UpstreamId: upstreamId}
+						return
+					}
 					out <- r
```
tonatoz: Same class of unguarded send one layer down (also the three `SubscribeTotalFailureError()` sends above). After the engine calls `terminate`, nothing drains `out`, so this forwarding goroutine can park on a full buffer and never observe `srcCtx.Done()`. Please guard these sends with `select { case out <- r: case <-srcCtx.Done(): return }`.

internal/upstreams/blocks/head_processor.go
```diff
@@ -140,11 +142,32 @@ func createHead(
...
+	id string, pollInterval time.Duration,
+	headMode config.HeadMode,
+	headConnector connectors.ApiConnector,
+	specific BlockChainSpecific,
+	options *chains.Options,
+) Head {
+	if headMode == config.HeadModePoll {
+		return NewRpcHead(ctx, id, options.InternalTimeout, pollInterval, specific)
+	}
+	if _, err := specific.SubscribeHeadRequest(); errors.Is(err, ErrUnsupportedHeadSubscriptions) {
```
tonatoz, line 166: The "any other error is a real failure and is left to the subscription head, which fails loudly on Start" contract does not hold for gRPC. `NewStream`/`SendMsg`/`CloseSend` do not wait for the server, so a node without `SubscriptionService` does not fail here: `Subscribe` returns successfully and `UNIMPLEMENTED` arrives later as an error frame. `SubscriptionHead.Start` therefore succeeds, the receive goroutine logs `got an error from heads subscription` and returns (`blocks/head.go:191`), the head never advances, `headNoUpdatesTimeout` fires, `OnNoHeadUpdates` does `Stop(); Start()`, and the cycle repeats indefinitely. The upstream stays headless with only a log line, and there is no path back to polling.

Combined with `HeadMode` defaulting to `subscribe` (`internal/config/defaults.go:336`), every existing Sui upstream on a fullnode without `SubscriptionService` changes behaviour on upgrade. Two options: default to `HeadModePoll`, or treat an `UNIMPLEMENTED` terminal frame on the first head subscription as a one-shot signal to switch to `NewRpcHead`.

Separately, this call discards the request and only classifies the error, while `NewSubHead` calls `SubscribeHeadRequest()` again on `Start`; a `SupportsHeadSubscriptions() bool` on `BlockChainSpecific` would express the capability check directly.

pkg/methods/method.go
```diff
@@ -124,7 +124,15 @@ func (m *Method) IsStickyCreate() bool {
 	return m.sticky.CreateSticky
 }
 
+// IsSubscribe reports whether the method is answered by a stream of frames
+// from one upstream: a JSON-RPC subscription, or either gRPC server-stream
+// shape. Both gRPC shapes take the subscription path through the flow (no
+// retries/hedges, SubscriptionRequestProcessor); only a clean EOF is read
+// differently, by GrpcCallType.
 func (m *Method) IsSubscribe() bool {
+	if m.grpcCallType.IsServerStream() {
```
tonatoz, line 133: This widening also reclassifies gRPC streams into `SubMethodGroup`, because `methodGroups.add` and `IsForceEnabled` key off `IsSubscribe()`. Result on this branch:

```
specs.GetSubMethods("sui") ->
[/sui.rpc.v2.SubscriptionService/SubscribeCheckpoints .../SubscribeTransactions .../SubscribeEvents
 /sui.rpc.v2.LedgerService/ListCheckpoints .../ListTransactions .../ListEvents]
```

That set becomes `chainSupervisor.GetChainState().SubMethods` (`internal/upstreams/chain_supervisor.go:89,344`), which is exactly the gate `subscribeMethodSupported` uses, and `supportsNativeSubscribeMethod` is now `IsSubscribeMethod` as well. So a dshackle client calling `NativeSubscribe(chain=sui, method="/sui.rpc.v2.SubscriptionService/SubscribeCheckpoints")` now passes both gates, and `emerald/grpc_blockchain.go:171` builds an `UpstreamJsonRpcRequest` whose JSON envelope is handed to `GrpcConnector.Subscribe` as the protobuf message body; before this change the mapping returned `UNIMPLEMENTED`. The streams are also advertised as dshackle subscribe methods now.

Keeping the group classification and the emerald gate on `Subscription.IsSubscribe` (a predicate distinct from "routed as a stream"), or rejecting non-`protocol.Grpc` request types in `mapNativeSubscribeMethod`, would close this.

tonatoz (changes requested):

internal/upstreams/flow/sub_processor.go
KirillPamPam: Fixed

internal/upstreams/flow/sub_aggregation.go
KirillPamPam: Fixed

internal/upstreams/blocks/head_processor.go
KirillPamPam: It's done deliberately, no need to overcomplicate it for now, as I think

pkg/methods/method.go
KirillPamPam: Everything will be handled later in the task of support grpc calls in Native* requests

tonatoz (approved):

PR #348, opened 2026-08-26
feat: add zstd support for http ingress and upstream connectors

mxssl:
## What

nodecore only ever spoke gzip on either edge. This adds zstd alongside it: the client-facing ingress negotiates a coding per request, and the upstream connectors offer zstd to nodes and decode whatever those nodes answer with. gzip and identity behave exactly as before.

zstd is both denser and much faster to decode than gzip, and node providers increasingly serve it. Against a live upstream that supports it, a 574KB `eth_getBlockByNumber` response goes out to the client as 114KB of zstd versus 137KB of gzip — 17% smaller, at a lower CPU cost per byte.

## `internal/compression` (new)

One package shared by both edges, so the codec pools exist once rather than per edge:

- `Negotiate(acceptEncoding)` — RFC 9110 §12.5.3 q-value parsing. Highest q wins, zstd breaks a tie, `q=0` is a refusal, anything unrecognised falls back to identity.
- `Offer` — the `zstd, gzip` nodecore advertises upstream.
- `WrapReader` / `AcquireWriter` — pooled encoders and decoders. Levels are pinned at the fastest setting of each codec, since a proxy pays compression on the critical path of every request, where CPU costs more than the last few percent of ratio.

Two limits worth calling out: encoders run at concurrency 1 (the library default spawns GOMAXPROCS goroutines *per encoder*, which at proxy concurrency is a goroutine count nobody asked for), and decoders cap the zstd window at 64MB (the library default is 64GB, so a hostile frame header can otherwise make you allocate on demand).

## Ingress

`compress.go` is generalized over codecs. It is re-ported from echo's current middleware rather than patched in place, because our copy predates a fix worth having: the status line is now held until the first body byte, so an empty response no longer ships a stale `Content-Encoding`. Echo's `MinLength` buffering is deliberately left out, and the `Level`/`Skipper` config went with the pool.

`decompress.go` replaces `middleware.Decompress()`, which only ever understood gzip. It decodes gzip and zstd request bodies, passes unknown codings through untouched, and answers 400 for a body that isn't the coding it claims to be. To keep that symmetric between codings, `WrapReader` checks the zstd frame magic eagerly — gzip already validated its header on reset, so zstd failing lazily mid-read would have been the odd one out.

## Upstream

Go's transport negotiates gzip on its own, but gzip is the only coding it knows, and it does so only while no `Accept-Encoding` is set. Offering zstd means setting the header explicitly, which turns that machinery off — so the connector now owns decoding for **both** codings.

- `applyConfigHeaders` sends `Accept-Encoding: zstd, gzip`, unless connector config pins one — an operator who pinned it has a reason, most likely a node that mishandles a coding.
- `dispatch` decodes before anything reads the body, so the buffered and streaming paths both see plain bytes. `decodedBody` ties the pooled codec's lifetime to the response body so neither path has to remember there are two things to close.
- A node answering with a coding we never offered is a partial failure rather than garbage handed to the client — the connector strips `Content-Encoding`, so those bytes would arrive unlabelled and unreadable.

## Two adjacent fixes

**`Content-Encoding` on decompressed requests.** It was left on the request after decoding and forwarded upstream, telling a node to decompress a body that is already plain. The middleware now deletes it.

**CORS clobbering `Vary`.** `setCorsHeaders` did `Set("Vary", "Origin")`, wiping the `Vary: Accept-Encoding` the compression middleware had just added. With two codings negotiable that is exactly how a shared cache ends up handing a zstd body to a gzip-only client. Changed to `Add`.

## Compatibility

No new config. Clients and upstreams that don't mention zstd see byte-identical behavior. One existing test changed contract: `TestRestRequest_HopByHopClientHeadersNotForwarded` asserted the upstream sees *no* `Accept-Encoding`, which is now the connector's own offer — the client's value still cannot survive, and the test says so explicitly.

gRPC and WebSocket are untouched: gRPC has its own compressor registry where zstd isn't a registered encoding, and WebSocket uses permessage-deflate. Neither speaks `Content-Encoding`.

## Testing

~20 new cases across 4 test files, written test-first. Full suite green (`go test` exit 0 over 93 packages), `-race` clean on the three touched packages, `go vet` clean.

Verified against a running binary, not just tests:

- **Live upstream that serves zstd:** 63.8KB on the wire decoded to 264KB; client-side negotiation returned zstd for `zstd` and for `gzip, zstd`, gzip for `gzip`, plain for no header, and every variant decoded to the same block.
- **Local node under load:** 1500 requests at 48-way concurrency across five coding combinations — 0 failures, 0 errors logged, every response body validated. This was the real test of decoder pooling.
- The node saw `Accept-Encoding: zstd, gzip` and received an 85-byte *plain*, unlabelled body even when the client sent a compressed one.
- Operator pin honored end to end: `gzip` and `identity` reached the node verbatim and both decoded correctly, with the client still getting zstd.
- WebSocket upgrade still returns 101 through the rewritten response writer, with no `Content-Encoding` on it.


mxssl:
## Review

Reviewed the ingress and upstream edges, then tested both — against unit tests, a running binary, and a live zstd-serving upstream. Baseline before any changes: `go build` and `go vet` clean, `go test -race -p 8 ./...` green over 93 packages (the two `caches/*_e2e` packages need Docker, unavailable in my environment).

The design is sound — independent negotiation per hop, stripping `Content-Encoding` after decoding, the connector owning decode once it sets `Accept-Encoding`, and `Vary` `Set`→`Add` are all right. I found five defects, three of which are reachable in production.

### 1. A pooled codec can be released while another goroutine is still reading it

`streamReadAhead` (`json_rpc_result_stream.go:54`) closes the stream **from the consuming goroutine specifically to unblock a producer parked in `Read`** — its own comment says so. That was safe when the thing being closed was `resp.Body`; `net/http` supports concurrent Close. It is not safe now that it closes a pooled decoder. `sync.Once` prevents a double release, not a release during a read.

This is not only a race. Reproduced under `-race`, the goroutine dump shows a hard deadlock:

```
goroutine A: zstd.(*Decoder).Reset -> drainOutput   [release path, blocked]
goroutine B: zstd.(*Decoder).Read  -> nextBlockSync [still in flight]
```

A gRPC client disconnecting mid-stream, or a failed `stream.Send`, hangs the cleanup goroutine. Both `json_rpc_result_stream.go:163` and `native_call_adapter.go:289` reach it.

**Fix:** `pooledReader` tracks in-flight reads; the codec is released by whichever of Read/Close finishes last, never under a live read. `decodedBody.Close` now closes the raw body *before* the decoder, since closing the raw body is what lets the parked read return.

### 2. `decoderMaxWindow` is a memory-amplification lever on the ingress

The 64MiB cap sits on a pool shared by **both** edges, so it also governs client request bodies. A zstd frame declares its own window and the decoder allocates it up front, so the sender picks how much memory nodecore commits.

Measured against a running binary, auth enabled, valid token:

```
24 requests, 962-byte bodies (23 KB total on the wire)
  heap_inuse before:    14 MB
  heap_inuse during: 1,394 MB
  heap_inuse after:  1,394 MB   <- pooled decoders retain their windows
```

Still 1.5 GB two minutes after traffic stopped. gzip cannot do this (fixed 32KiB window), and echo's `Decompress()` ignored zstd entirely, so the lever is new.

**Fix:** cap at 8MiB. RFC 9659 §3 makes that the normative limit for this content coding — *"decoders MUST support a Window_Size of up to and including 8 MB, and encoders MUST NOT generate frames requiring a Window_Size larger than 8 MB"*. Every mainstream setting still fits: `zstd -1/-3/-9/-19` declare 0.5/2/4/8 MiB, klauspost tops out at 8MiB. Same attack after the fix: **12 MB → 13 MB**, with a clean 400.

(The comment's "library defaults to 64GiB" conflated two settings — the window default is `MaxWindowSize`, 512MiB; 64GiB is `maxDecodedSize`. Corrected.)

### 3. The response writer keeps serving a codec it has already released

`compress.go` restores `res.Writer` only when nothing was written, but always closes and releases the encoder — so after a write, `res.Writer` still points at a released codec.

Echo's `DefaultHTTPErrorHandler` bails when `Response().Committed`, which covers ordinary writes. **A rejected WebSocket upgrade does not commit**: gorilla writes its 400 through `c.Response().Writer` (`http_server.go:139`), bypassing echo's bookkeeping. The error handler then writes through an encoder that may already belong to another response. Confirmed with a real rejected handshake.

**Fix:** writes after release are refused; `Flush` becomes a no-op. The client now gets gorilla's complete, decodable 400 and nothing reaches the pooled encoder.

### 4. An empty body is an error under gzip and a success under zstd

Compared against the PR base (`6233645`):

| upstream response | base | this PR |
|---|---|---|
| 200, empty body, `Content-Encoding: gzip` | ok, empty result | **partial failure** |
| 204, `Content-Encoding: gzip` | ok, empty result | **partial failure** |
| 200, empty body, `Content-Encoding: zstd` | garbage passed through | ok, empty result |
| `Content-Encoding: br` | raw br bytes as the result | partial failure ✅ |
| `Content-Encoding: zstd` | raw zstd bytes as the result | decoded ✅ |

Base didn't fail because Go's transparent `gzipReader` surfaces an empty stream as `io.EOF`, which `io.ReadAll` reads as a clean end. Echo's decompress middleware went out of its way to allow it too (`if err == io.EOF { //ignore if body is empty }`). The eager `Reset` turns both into errors — and the two codings now disagree on identical input, which is the asymmetry `checkZstdMagic` exists to prevent.

**Fix:** an empty body is the empty payload it is, under both codings, on both edges. Live before/after on the same request: `"invalid compressed request body"` → passes through to routing.

### 5. `Negotiate` lets `*` speak for codings the client named itself

`*` maps to zstd unconditionally, so `Accept-Encoding: zstd;q=0, *` returns **zstd** — to precisely the client that just said it cannot read zstd. RFC 9110 §12.5.3: `*` matches only codings *not* explicitly listed. Also `zstd;q=0.1, gzip;q=0.5, *` → zstd, and an explicit `identity;q=1, gzip;q=0.5` was ignored.

**Fix:** the wildcard covers only unmentioned codings, an explicit identity preference is honoured, `q` is clamped to the `[0,1]` the grammar defines (`ParseFloat` accepts `1e9`, `+Inf`, `NaN`), and repeated `Accept-Encoding` field lines are read together rather than first-only. Verified live:

```
zstd;q=0, *                 -> gzip
*, zstd;q=0                 -> gzip
zstd;q=0.1, gzip;q=0.5, *   -> gzip
identity;q=1, gzip;q=0.5    -> (none)
gzip, deflate, br, zstd     -> zstd
```

### What held up

Against a live zstd-serving upstream (472KB plain → 78KB zstd vs 89KB gzip), through a logging relay so both hops were visible:

- All 12 client `Accept-Encoding` variants negotiated correctly; every body decoded to the same block. Client edge: **zstd 130,126 B vs gzip 147,656 B** on an 823KB body, 11.9% smaller.
- Across all 8 two-hop combinations the node saw `Accept-Encoding: zstd, gzip` every time and received the forwarded body **plain and unlabelled**, even when the client sent gzip or zstd — the `Content-Encoding` deletion fix works end to end.
- CORS fix confirmed: `Vary: Accept-Encoding` survives alongside `Vary: Origin`.
- WebSocket upgrade returns 101 with no `Content-Encoding` under every `Accept-Encoding`.
- **5,120 requests at 64-way concurrency**, mixed codings, every body validated: 0 failures, goroutines stable, no leak. Pooled encoder cost is a non-issue — zstd 859KiB vs gzip 795KiB per live encoder.

### Not addressed here

**No decompressed-size cap.** A 13KB body still expands to 64MB (zstd 5040:1, gzip 1011:1), and there is no request body limit anywhere in the repo. That needs a size policy rather than a compression change — worth a separate issue.

**No 406 for `identity;q=0`.** Serving a plain body is what an RPC client can use, and RFC 9110 permits disregarding negotiation when nothing is acceptable. Documented in the code as deliberate.

**Unrelated:** `TestGrpcConnectorParentDeadlineIsTotalFailure` is flaky — it asserts a 100ms parent deadline beats the connector's own, and fails 2 of 5 runs under CPU load **on the stock branch with no changes applied**. Not caused by this PR, but worth fixing separately.

---

Fixes for all five are committed (`117a2a9`), with a test per defect — each verified to fail on the unpatched code and pass on the patched code. `go vet` clean, 93 packages green under `-race`, and the live checks above re-run against the fixed binary.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

mxssl:
## Client interoperability testing

Ran real HTTP stacks and web3 libraries from five languages against nodecore's client edge, with a pass-through proxy in front recording exactly what each client sent and what went back on the wire. **59 cases, 0 failures.**

Responses were a real 512 KB mainnet block (338 txs), replayed locally so the numbers are comparable across clients:

```
plain 511,288 B   gzip 103,185 B (20.2%)   zstd 91,492 B (17.9%)
zstd is 11.3% smaller than gzip
```

### What each library asks for by default

This is the part worth knowing, because almost nothing opts into zstd on its own:

| Library | Default `Accept-Encoding` | Served | Wire |
|---|---|---|---|
| Go `net/http` / **go-ethereum** | `gzip` | gzip | 103 KB |
| **web3.py**, requests, httpx *(with `zstandard`)* | `gzip, deflate, br, zstd` | **zstd** | 91 KB |
| **web3.py**, requests, httpx *(without)* | `gzip, deflate` | gzip | 103 KB |
| Node `fetch` (undici) | `gzip, deflate` | gzip | 103 KB |
| node-fetch | `gzip, deflate, br` | gzip | 103 KB |
| axios | `gzip, compress, deflate, br` | gzip | 103 KB |
| **ethers v6** | `gzip` | gzip | 103 KB |
| **viem** | `gzip, deflate` | gzip | 103 KB |
| **web3.js v4** | `gzip,deflate` | gzip | 103 KB |
| Rust reqwest *(gzip+zstd features)* | `zstd,gzip` | **zstd** | 91 KB |
| **alloy** | *(none sent)* | identity | 511 KB |
| Java OkHttp / **web3j** | `gzip` | gzip | 103 KB |
| Java JDK `HttpClient` | *(none sent)* | identity | 511 KB |

Only two stacks reach for zstd unprompted: Rust's reqwest, and the Python stack when `zstandard` is installed. Everything else lands on gzip and is byte-for-byte unaffected by this PR. Negotiation degrades exactly as intended — remove `zstandard` from the Python environment and requests/httpx/web3.py drop straight back to gzip.

Two clients send no `Accept-Encoding` at all (alloy, JDK `HttpClient`) and correctly get an uncompressed body. Worth noting that alloy users get no compression from nodecore today — bare `reqwest::Client::new()` with the same features advertises `zstd,gzip`, so alloy is suppressing it, not the feature flags.

### Explicit codings

Every client was also driven with `Accept-Encoding: gzip`, `zstd`, and `gzip, zstd`, and with **gzip- and zstd-compressed request bodies**. All correct: the right coding came back, every body decoded to the same block, and every compressed request body was decoded and reached the handler as plain JSON.

Three behaviours are worth calling out, none of them nodecore bugs:

- **node-fetch v3 cannot decode zstd.** Ask it for `zstd` (or `gzip, zstd`) and the body reaches the caller still encoded — it decodes correctly by hand. Not reachable via defaults; only an app that hand-writes `Accept-Encoding` to include a coding its client can't handle gets here, which is the client lying about what it accepts.
- **OkHttp and Go's transport turn off transparent decompression when you set `Accept-Encoding` yourself.** Both hand you raw bytes, by design. Caught my own harness out first.
- **The JDK's `HttpClient` never decompresses anything**, gzip included, and advertises nothing. Unchanged by this PR.

### Against the live upstream

Repeated with nodecore pointed at a real mainnet endpoint that serves zstd, with a relay recording the upstream hop:

```
upstream hop:  Accept-Encoding: zstd, gzip  ->  Content-Encoding: zstd  (89,471 B)
client hop:    each library got the coding it asked for
```

web3.py, ethers v6, viem, go-ethereum, web3j and alloy each fetched a full latest block through the proxy and parsed it. Live upstream figures: 472 KB plain → 88.9 KB gzip → 78.0 KB zstd, **12.3% smaller than gzip**.

### Versions

Go 1.27.1 / go-ethereum v1.17.5 · Python 3.12 / requests 2.34.2, httpx 0.28.1, urllib3 2.7.0, web3.py 8.0.0, zstandard 0.25.0 · Node 24.21.0 / undici 8.10.2, node-fetch 3.3.2, axios 1.20.0, ethers 6.17.0, viem 2.56.3, web3.js 4.16.0 · Rust 1.98.1 / reqwest 0.12.28, alloy 1.8.3 · Temurin 21 / OkHttp 4.12.0, web3j 4.12.2

mxssl:
## Review follow-ups

Three findings from a review of this PR, fixed in da769cc. Each has a test that fails without the fix.

**A cancelled request was charged to the upstream** — `internal/upstreams/connectors/http_connector.go`. Both codings read the head of the response body to validate it (gzip parses its header, zstd peeks its frame magic), so a client that walks away mid-response fails right there. That was reported as a *partial* failure, which `IsRetryable` answers true to: a healthy node took the error-rate penalty and the request was re-dispatched with nobody left to answer it. It is a context failure, like the ones the branches either side of it already report.

**The ingress pinned an encoder for the life of a WebSocket connection** — `internal/server/http_server/compress.go`. The encoder was taken when the middleware wrapped the response, not at the first body byte. A WebSocket upgrade hijacks the connection out from under that writer and then lives as long as the socket does, so every ws connection held an encoder it never wrote a single byte through. It is taken in `commit()` now, which also spares every bodyless response (204, 304, a rejected upgrade) a pool round-trip, and a failed acquire degrades to serving the body plain rather than dropping the wrapper.

**Request bodies decoded without a size limit** — `internal/server/http_server/decompress.go`. The ratio is the sender's to choose: DEFLATE tops out near 1000:1, zstd has no comparable ceiling. Decoded bodies are now capped at 32MiB — far above any real JSON-RPC batch — and a body past the cap fails rather than being handed to the handler silently short. A coding nodecore does not decode is left uncapped: those are bytes the client had to send in full.

## Verification

`go test ./...` green except the two docker-backed cache e2e packages (no docker on this machine); `-race` clean on the three touched packages; `go vet` and `gofmt` clean.

### Five client languages against live upstreams

Five nodecore instances against four live upstreams — three that serve zstd on both edges, one pinned by connector config to `Accept-Encoding: identity` so the node answers plain while the client still gets zstd — plus a beacon-chain REST upstream.

Clients in **Rust, Java, Python, Go and JavaScript**, each speaking HTTP directly rather than through a library that negotiates `Accept-Encoding` and decodes the body on its own, since that is the behaviour under test. Per client, per target: `Accept-Encoding: zstd`, `gzip`, `gzip, zstd`, absent and `br`, plus request bodies sent under `Content-Encoding: zstd` and `gzip`. Every response was decoded by the client and validated against the same block (hash + 67 transactions) or a beacon header.

**165 cases (33 × 5 languages), all passing.** Byte counts agree exactly across all five languages:

| Accept-Encoding | Content-Encoding | wire | decoded |
|---|---|---|---|
| `zstd` | zstd | 63,235 | 340,008 |
| `gzip` | gzip | 70,882 | 340,008 |
| `gzip, zstd` | zstd | 63,235 | 340,008 |
| absent | — | 340,008 | 340,008 |
| `br` | — | 340,008 | 340,008 |
| `zstd` + request body in zstd | zstd | 63,235 | 340,008 |
| `zstd` + request body in gzip | zstd | 63,235 | 340,008 |

The same matrix run against the pre-fix build is byte-identical, so nothing about negotiation moved.

### Before / after, per fix

**Cancelled decodes.** Against a node that sends response headers and then stalls, five client requests aborted mid-decode:

| | log lines | `nodecore_upstream_errors_total` |
|---|---|---|
| before | 5 × `cannot decode the upstream response body` | 5 |
| after | none | metric never incremented |

**WebSocket encoder pinning.** Live heap after a forced GC, holding N ws connections open with `Accept-Encoding: zstd` on the handshake:

| connections | before | after |
|---|---|---|
| 200 | 125.9 MiB (644.6 KiB/conn) | 3.0 MiB (15.4 KiB/conn) |
| 1000 | 629.6 MiB (644.7 KiB/conn) | 15.2 MiB (15.5 KiB/conn) |

**Decoded-body cap.** One request, 16,485 bytes on the wire, 512MiB decoded (32,567:1), sent as a valid JSON-RPC body with a large padding field:

| | status | allocated | peak heap | time |
|---|---|---|---|---|
| before | 200, served | 2,254 MiB | 3,293 MiB | 0.72s |
| after | 400, refused | 78 MiB | 124 MiB | 0.03s |

WebSocket upgrades still return 101 with no `Content-Encoding` on either build.

### Unrelated finding

One of the beacon-chain endpoints tested answers `400 Unsupported method` to `GET /eth/v1/node/health`, which `BeaconChainHealthValidator` requires, so nodecore marks that upstream unavailable and answers `no available upstreams` to every request — while `/eth/v1/beacon/headers` and `/eth/v1/node/syncing` on the same endpoint both work. Nothing to do with this PR, but it makes that provider's beacon endpoint unusable as a nodecore upstream today.

KirillPamPam (approved):

PR #349, opened 2026-08-26
lower bounds: never trust a cached bound alone; live-correct STATE from pruned errors

l0gun0v:
## Problem

coreth full nodes (avalanche, flare) keep state for the last ~32 blocks but persist the trie of every `commit-interval` (4096) block forever (`core/state_manager.go`: `Commit` on `number % commitInterval == 0`, older roots dereferenced after `state-history`). `detectPlain` confirmed its cached bound with a single probe, so once a search landed on such a block the bound was republished unchanged every 180 s while the head moved on. All our avalanche/flare upstreams sat 90k–950k blocks behind head (`bound % 4096 == 0`, one value for 24 h+), and dproxy routed archive-depth `eth_call` to them: 15k–72k `missing trie node` errors/day to clients (Zendesk #3423).

## Change

- `lower_bound_search.go`: no cached-bound shortcut in `detectPlain`; the cached bound stays the left edge of the search (offset search keeps it, same as dshackle).
- `request_processor.go`: pruned-history errors on `eth_call`, `eth_getBalance`, `eth_getStorageAt`, `eth_getCode` now raise the STATE bound — only proof/logs/trace were mapped before, so STATE was never corrected live. (`eth_estimateGas` / `eth_getTransactionCount` have no tag parser in the spec.)
- `LowerBoundUpstreamStateEvent`: a lower value never replaces an existing bound unless it is the archive reset (1); previously `AddLowerBound` overwrote a live correction with the next periodic publish.

## Verification

- Unit: swiss-cheese probe (state at `%4096==0` + last 32 blocks, moving head) — bound follows `head-32` every cycle; live handler for state methods incl. blockRef params; no-downgrade event rule. `go test -race ./...` green (except Docker-only cache e2e).
- Local E2E vs `origin/main` on a mock coreth (head parked so the first search lands on root 8192, then +1 block/s): old binary reports `8192, 8192, 8192`; this branch `8192 → 16503 → 16683` (= head−32 each cycle).
- Live path on the gRPC `SubscribeChainStatus` stream (what dproxy consumes): `8192 → 9001` right after `eth_call@9000` fails, `→ 16001` after `eth_getBalance@16000`, unchanged after a successful call on root 8192, `→ 16502` on the next periodic cycle.


internal/upstreams/flow/request_processor.go
```diff
@@ -300,7 +300,20 @@ func liveLowerBoundFromPrunedError(method string, param specs.MethodParam, respo
...
+// eth_getTransactionCount have no tag parser in the spec, so they can't be used.
+var stateMethods = map[string]struct{}{
+	"eth_call":         {},
+	"eth_getBalance":   {},
+	"eth_getStorageAt": {},
+	"eth_getCode":      {},
+}
+
 func lowerBoundTypeForMethod(method string) (protocol.LowerBoundType, bool) {
+	if _, ok := stateMethods[method]; ok {
```
KirillPamPam, line 314: I just realised one thing - there is the `isPrunedHistoryError` that check if an error message contains one of the errors from the list. But there is the thing - there might be "execution reverted" errors that also may contains such errors that may cause issues. So I suggest to ignore "execution reverted"  at all

internal/protocol/upstream_state_events.go
```diff
@@ -34,6 +34,12 @@ func (l *LowerBoundUpstreamStateEvent) Same(state UpstreamState) bool {
 }
 
 func (l *LowerBoundUpstreamStateEvent) ProcessEvent(state UpstreamState) UpstreamState {
+	// Bounds only move up (or reset to 1 for a fully archival node) — same rule the periodic
+	// detector applies. Without it the detector's next publish would undo a live correction made
+	// from a pruned-history error (request_processor.liveLowerBoundFromPrunedError).
+	if existing, ok := state.LowerBoundsInfo.GetLowerBound(l.Data.Type); ok && l.Data.Bound != 1 && l.Data.Bound < existing.Bound {
```
KirillPamPam: I suggest to move this check in the method `Same` where we should have such things. Additionally:

1. This is a copy of the mechanism that we have in the LowerBoundProcessor and it's much better to update a lower bound through it, but it takes a lot of work to refactor this code. So I suggest just to move this check in `Same()` with a comment and a task that it should be refactored via LowerBoundProcessor
2. It's also necessary to add another check here so far - the bound must not be greater then the head value.

KirillPamPam (approved):

PR #354, opened 2026-08-28
Support grpc requests in NativeCall

KirillPamPam:
# NativeCall: serve native gRPC calls via `grpc_data` items

## Summary

The dshackle `Blockchain.NativeCall` stream now accepts native unary gRPC calls (Sui
`sui.rpc.v2.*`) alongside JSON-RPC and REST items. A `NativeCallItem` with the new
`grpc_data` oneof branch is routed through the same execution flow as the gRPC chain
ingress and answered with one buffered reply item carrying the response message bytes,
the upstream's headers **and trailers**, and — on failure — a canonical gRPC status.

Bytes-only throughout: request and response messages are never parsed.

Depends on `github.com/drpcorg/public` **v1.1.0** (bumped here), which added:

```proto
message NativeCallItem { oneof data { ...; GrpcData grpc_data = 9; } }
message GrpcData { bytes payload = 1; repeated KeyValue metadata = 2; }
message NativeCallReplyItem { ...; repeated KeyValue response_trailers = 16; }
```

## Contract for a gRPC item

**Request** — `method` is the full gRPC method name (`/sui.rpc.v2.LedgerService/GetObject`),
`grpc_data.payload` the serialized request message (no 5-byte wire frame prefix),
`grpc_data.metadata` the call metadata to forward. Credential metadata (`x-nodecore-key`,
`x-nodecore-token`, `authorization`) is stripped before forwarding, exactly like the ingress does.
`chunk_size` is ignored: a unary gRPC reply is always one unchunked message.

**Success** — `payload` = response message bytes verbatim, `signature` over those bytes when
a nonce is given, `response_headers` / `response_trailers` = upstream metadata.

**Error** — existing fields, gRPC vocabulary:

| field | value |
|---|---|
| `item_error_code` | canonical gRPC code (0–16) — never nodecore's internal `ResponseError.Code` |
| `error_message` | the status message |
| `error_as_is` | serialized `google.rpc.Status` when the upstream attached typed details; empty otherwise |
| `error_data` | always empty for gRPC items |
| `response_headers` / `response_trailers` | present on error items too (e.g. `RESOURCE_EXHAUSTED` rate-limit hints ride in trailers) |

An upstream status is replayed verbatim; nodecore's own errors (no upstreams, rate limit,
unknown method, timeout, auth, …) are mapped onto the 17-code model **in nodecore**, so the
client needs no mapping table of its own: `status.FromProto(error_as_is)` when set, else
`status.New(item_error_code, error_message)`.

Rejected up front (no upstream involved): unknown method → `UNIMPLEMENTED`; a server-stream
spec method → `INVALID_ARGUMENT` pointing at `NativeSubscribe`; missing `grpc_data` →
`INVALID_ARGUMENT`.

## Changes

- **`protocol.GrpcStatusOf(*ResponseError) *status.Status`** — the nodecore-error → gRPC
  status mapping, moved out of `grpc_ingress/chain_ingress.go` (`grpcStatusFromResponseError`)
  so the ingress and the emerald server share one table. The ingress behaviour is unchanged.
- **`internal/server/emerald`**
  - `native_call_adapter.go` is split per API kind: `native_call_adapter_jsonrpc.go`,
    `native_call_adapter_rest.go`, `native_call_adapter_grpc.go` (new adapter); the shared
    file keeps the `nativeCallAdapter` interface, `adapterFor` and the reply machinery.
  - `nativeCallAdapter` gains `ErrorItem(requestID, err)`, so pre-dispatch failures (invalid
    payload, selector conflicts, signature requested without a signer) render in the item's
    own error vocabulary. The adapter is now chosen before the signing check.
  - `sendReply` takes an `errorItemRenderer` — nodecore codes + `error_data` for JSON-RPC/REST,
    canonical status for gRPC — and reads response metadata via the
    `HasResponseHeaders`/`HasResponseTrailers` capabilities instead of a
    `*GenericUpstreamResponse` type assertion, so metadata on `*ReplyError` (retryable upstream
    errors) is no longer dropped. Trailers are stamped on success, error and first stream chunk.
  - New `replyMeta` (request id, upstream id/version, finalization, headers, trailers) built
    once per response and `stamp`ed onto items; `streamNativeCallBody` goes from 11 positional
    params to `(stream, reader, mode, hint, meta)`.
- **`go.mod`** — `github.com/drpcorg/public v1.0.0 → v1.1.0`.

JSON-RPC and REST items are unaffected apart from the (previously empty) trailers field and
the `*ReplyError` headers that now ride along.

## Out of scope (follow-ups)

- `NativeSubscribe` for gRPC server-stream methods (still JSON-RPC only).
- Quorum for gRPC items — quorum parameters are currently read only from the HTTP query
  string, so quorum is unreachable via `NativeCall` for any item kind today.

## Testing

- `internal/protocol/grpc_status_of_test.go` — verbatim replay with/without details, every
  nodecore code mapping, nil input.
- `internal/server/emerald/grpc_blockchain_test.go` — gRPC item build (method, body, sanitized
  metadata, `chunk_size` ignored), unknown / server-stream / missing-data / signing-unavailable
  items, success item with headers + trailers, upstream status with details round-tripping
  through `error_as_is`, status without details, nodecore error → `UNAVAILABLE`, trailers on
  `*ReplyError`, JSON-RPC regression (nodecore codes kept, trailers empty).
- `go test -race ./...` green, `golangci-lint` clean.

### DIFF
--- a/internal/server/emerald/grpc_blockchain.go
+++ b/internal/server/emerald/grpc_blockchain.go
@@ -4,7 +4,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"net/http"
 	"strconv"
 	"strings"
 	"time"
@@ -71,18 +70,18 @@ func (s *GrpcBlockchainService) NativeCall(request *dshackle.NativeCallRequest, 
 		return err
 	}
 	if request == nil {
-		return stream.Send(nativeCallErrorItem(0, protocol.ClientError(fmt.Errorf("request is nil")), flow.NoUpstream, nil, nil))
+		return stream.Send(noUpstreamErrorItem(0, protocol.ClientError(fmt.Errorf("request is nil"))))
 	}
 	if s.appCtx == nil || s.appCtx.UpstreamSupervisor == nil {
-		return stream.Send(nativeCallErrorItem(0, protocol.NoAvailableUpstreamsError(), flow.NoUpstream, nil, nil))
+		return stream.Send(noUpstreamErrorItem(0, protocol.NoAvailableUpstreamsError()))
 	}
 
 	configuredChain, chainSupervisor := s.resolveChain(request.GetChain())
 	if configuredChain == nil {
-		return stream.Send(nativeCallErrorItem(0, protocol.WrongChainError(strconv.Itoa(int(request.GetChain()))), flow.NoUpstream, nil, nil))
+		return stream.Send(noUpstreamErrorItem(0, protocol.WrongChainError(strconv.Itoa(int(request.GetChain())))))
 	}
 	if chainSupervisor == nil {
-		return stream.Send(nativeCallErrorItem(0, protocol.NoAvailableUpstreamsError(), flow.NoUpstream, nil, nil))
+		return stream.Send(noUpstreamErrorItem(0, protocol.NoAvailableUpstreamsError()))
 	}
 
 	requests, items, preResponses := s.buildNativeCallRequests(configuredChain, request)
@@ -118,7 +117,9 @@ func (s *GrpcBlockchainService) NativeCall(request *dshackle.NativeCallRequest, 
 			// Without the originating item we don't know its nonce, so a success
 			// here could silently drop a signature the client asked for.
 			log.Warn().Msgf("no request found for id %s, cannot build a reply", wrapper.RequestId)
-			if err := stream.Send(nativeCallErrorItem(parseCallItemID(wrapper.RequestId), protocol.ServerError(), wrapper.UpstreamId, nil, nil)); err != nil {
+			replyItem := nativeCallErrorItem(parseCallItemID(wrapper.RequestId), protocol.ServerError(), nil)
+			replyItem.UpstreamId = wrapper.UpstreamId
+			if err := stream.Send(replyItem); err != nil {
 				return err
 			}
 			continue
@@ -261,12 +262,12 @@ func (s *GrpcBlockchainService) buildNativeCallRequests(
 	preResponses := make([]*dshackle.NativeCallReplyItem, 0)
 
 	for _, item := range request.GetItems() {
+		adapter := adapterFor(item)
 		if err := signingUnavailable(item.GetNonce(), s.signer); err != nil {
 			log.Warn().Msgf("item %d requested a signature but response signing is not configured", item.GetId())
-			preResponses = append(preResponses, nativeCallErrorItem(item.GetId(), protocol.ServerErrorWithCause(err), flow.NoUpstream, nil, nil))
+			preResponses = append(preResponses, adapter.ErrorItem(item.GetId(), protocol.ServerErrorWithCause(err)))
 			continue
 		}
-		adapter := adapterFor(item)
 		builtRequest, failure := adapter.BuildRequest(configuredChain, item, request.GetSelector(), request.GetChunkSize())
 		if failure != nil {
 			preResponses = append(preResponses, failure)
@@ -283,18 +284,11 @@ func (s *GrpcBlockchainService) buildNativeCallRequests(
 // buffered response is always a single unchunked item however large: chunk_size
 // selects a streaming request, it is not a framing size for in-memory payloads.
 // Chunked replies come from streamNativeCallBody instead.
-func nativeCallSuccessItem(
-	requestID uint32,
-	upstreamID string,
-	payload []byte,
-	headers http.Header,
-) *dshackle.NativeCallReplyItem {
+func nativeCallSuccessItem(requestID uint32, payload []byte) *dshackle.NativeCallReplyItem {
 	return &dshackle.NativeCallReplyItem{
-		Id:              requestID,
-		Succeed:         true,
-		Payload:         payload,
-		UpstreamId:      upstreamID,
-		ResponseHeaders: mapHeaders(headers),
+		Id:      requestID,
+		Succeed: true,
+		Payload: payload,
 	}
 }
 
@@ -361,24 +355,16 @@ func (e *nativeCallChunkEmitter) Finish() error {
 	return e.WriteChunk(nil, true)
 }
 
-func nativeCallErrorItem(
-	requestID uint32,
-	responseError *protocol.ResponseError,
-	upstreamID string,
-	errorAsIs []byte,
-	headers http.Header,
-) *dshackle.NativeCallReplyItem {
+func nativeCallErrorItem(requestID uint32, responseError *protocol.ResponseError, errorAsIs []byte) *dshackle.NativeCallReplyItem {
 	if responseError == nil {
 		responseError = protocol.ServerError()
 	}
 
 	replyItem := &dshackle.NativeCallReplyItem{
-		Id:              requestID,
-		Succeed:         false,
-		ErrorMessage:    responseError.Message,
-		ItemErrorCode:   int32(responseError.Code),
-		UpstreamId:      upstreamID,
-		ResponseHeaders: mapHeaders(headers),
+		Id:            requestID,
+		Succeed:       false,
+		ErrorMessage:  responseError.Message,
+		ItemErrorCode: int32(responseError.Code),
 	}
 	if responseError.Data != nil {
 		replyItem.ErrorData = nativeCallErrorData(responseError.Data)
@@ -390,7 +376,14 @@ func nativeCallErrorItem(
 	return replyItem
 }
 
-func mapHeaders(headers http.Header) []*dshackle.KeyValue {
+// noUpstreamErrorItem is a failure raised before any upstream was involved.
+func noUpstreamErrorItem(requestID uint32, responseError *protocol.ResponseError) *dshackle.NativeCallReplyItem {
+	replyItem := nativeCallErrorItem(requestID, responseError, nil)
+	replyItem.UpstreamId = flow.NoUpstream
+	return replyItem
+}
+
+func mapHeaders[M ~map[string][]string](headers M) []*dshackle.KeyValue {
 	keyValueHeaders := make([]*dshackle.KeyValue, 0, len(headers))
 	for key, values := range headers {
 		for _, value := range values {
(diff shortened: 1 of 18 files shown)


internal/server/emerald/grpc_blockchain.go
```diff
@@ -71,18 +70,18 @@ func (s *GrpcBlockchainService) NativeCall(request *dshackle.NativeCallRequest,
...
 	}
 	if s.appCtx == nil || s.appCtx.UpstreamSupervisor == nil {
-		return stream.Send(nativeCallErrorItem(0, protocol.NoAvailableUpstreamsError(), flow.NoUpstream, nil, nil))
+		return stream.Send(noUpstreamErrorItem(0, protocol.NoAvailableUpstreamsError()))
 	}
 
 	configuredChain, chainSupervisor := s.resolveChain(request.GetChain())
 	if configuredChain == nil {
-		return stream.Send(nativeCallErrorItem(0, protocol.WrongChainError(strconv.Itoa(int(request.GetChain()))), flow.NoUpstream, nil, nil))
+		return stream.Send(noUpstreamErrorItem(0, protocol.WrongChainError(strconv.Itoa(int(request.GetChain())))))
```
tonatoz: Request-level failures are rendered in the nodecore vocabulary for every item kind, and those codes collide with the canonical gRPC ones: `internal/protocol/errors.go:8-13` defines `NoAvailableUpstreams=1, WrongChain=2, CtxErrorCode=3, SubscribeTotalFailure=4, NoApiConnectors=5`, i.e. exactly `CANCELLED/UNKNOWN/INVALID_ARGUMENT/DEADLINE_EXCEEDED/NOT_FOUND`.

A client that sent only `grpc_data` items and follows the contract in the PR description (`status.New(item_error_code, error_message)`) reads a wrong-chain reply as `UNKNOWN(2)` and a no-upstreams reply as `CANCELLED(1)`. The failure is silent — nothing marks the item as being in a different code space.

`request.GetItems()` is already available here, so these four early returns could render one failure per item through `adapterFor(item)`. That also fixes the second half of the problem: `id: 0` matches no item the client sent, so it cannot correlate the failure.

There is currently no test covering this path — the new cases all call `buildNativeCallRequests` or `SendReply` directly, never `NativeCall` end to end.

internal/server/emerald/grpc_blockchain.go
```diff
@@ -118,7 +117,9 @@ func (s *GrpcBlockchainService) NativeCall(request *dshackle.NativeCallRequest,
 			// Without the originating item we don't know its nonce, so a success
 			// here could silently drop a signature the client asked for.
 			log.Warn().Msgf("no request found for id %s, cannot build a reply", wrapper.RequestId)
-			if err := stream.Send(nativeCallErrorItem(parseCallItemID(wrapper.RequestId), protocol.ServerError(), wrapper.UpstreamId, nil, nil)); err != nil {
+			replyItem := nativeCallErrorItem(parseCallItemID(wrapper.RequestId), protocol.ServerError(), nil)
```
tonatoz, line 126: Same code-space issue as the early returns above: a gRPC item reaching this branch gets `item_error_code = 500`, which is outside the 0-16 range the gRPC contract promises. The item kind is genuinely unknown here (the `items` lookup is what failed), so either the per-item rendering suggested for the early returns covers this too, or the exception should be documented as part of the contract.

internal/server/emerald/native_call_adapter_grpc.go
```diff
@@ -0,0 +1,93 @@
...
+func grpcNativeCallErrorItem(requestID uint32, responseError *protocol.ResponseError, _ []byte) *dshackle.NativeCallReplyItem {
+	grpcStatus := protocol.GrpcStatusOf(responseError)
+	replyItem := &dshackle.NativeCallReplyItem{
+		Id:            requestID,
+		Succeed:       false,
+		ErrorMessage:  grpcStatus.Message(),
+		ItemErrorCode: int32(grpcStatus.Code()),
+	}
+	if upstreamStatus, ok := protocol.GrpcStatusFromError(responseError); ok && len(upstreamStatus.StatusProto) > 0 {
+		replyItem.ErrorAsIs = append([]byte(nil), upstreamStatus.StatusProto...)
```
tonatoz: `ErrorAsIs` is set whenever `StatusProto` is non-empty, but `GrpcStatusOf` silently falls back to code/message when those bytes fail to unmarshal. In that case the item carries a status the client cannot parse while the contract tells it to prefer `status.FromProto(error_as_is)`.

Setting `ErrorAsIs` only when the bytes unmarshal keeps the two fields consistent. Since `GrpcStatusOf` already unmarshals them, returning that information (or unmarshalling once here) avoids a second pass.

internal/server/emerald/native_call_adapter.go
```diff
@@ -246,40 +184,31 @@ func nativeCallFinalizationData(wrapper *protocol.ResponseHolderWrapper) *dshack
...
 		item := &dshackle.NativeCallReplyItem{
-			Id:         requestID,
+			Id:         meta.requestID,
 			Succeed:    true,
 			Payload:    chunk,
 			Chunked:    true,
 			FinalChunk: final,
 		}
 		// Response-level metadata travels on the first chunk only.
 		if first {
```
tonatoz, line 193: Trailers are read in `newReplyMeta` before the body is consumed and stamped on the first chunk. gRPC trailers arrive after the last message, so for any streamed gRPC response they would always be empty.

Unreachable today (server-stream methods are rejected in `grpcNativeCallAdapter.BuildRequest` and a unary reply never streams), but it is a trap for the `NativeSubscribe` follow-up. Stamping trailers on the final chunk instead, or at least noting the constraint here, would keep that follow-up honest.

internal/server/emerald/native_call_adapter.go
```diff
@@ -219,19 +114,62 @@ func sendReply(
...
+	item.Finalization = m.finalization
+	item.ResponseHeaders = m.headers
+	item.ResponseTrailers = m.trailers
+	return item
+}
+
+// responseMetadata reads the upstream's response metadata through the optional
+// capabilities, so success responses and error replies (a RESOURCE_EXHAUSTED
+// with rate-limit hints in its trailers is a *ReplyError) are treated alike.
+func responseMetadata(response protocol.ResponseHolder) (http.Header, map[string][]string) {
```
tonatoz: This is the third copy of the same capability read: `flow.subResponseMetadata` (`internal/upstreams/flow/sub_processor.go:174-183`) is identical, and `grpc_ingress.forwardResponseMetadata` (`internal/server/grpc_ingress/chain_ingress.go:218-227`) is the same two type assertions inlined. A single `protocol.ResponseMetadata(ResponseHolder) (http.Header, map[string][]string)` next to the interface definitions would consolidate them — the same move already made for `GrpcStatusOf` in this PR.

internal/server/emerald/native_call_adapter.go
```diff
@@ -36,12 +32,25 @@ type nativeCallAdapter interface {
 		nonce uint64,
 		signer signature.ResponseSigner,
 	) error
+
+	// ErrorItem renders a failure that happened before or outside the flow
+	// (no upstream) in the item's own error vocabulary.
+	ErrorItem(requestID uint32, responseError *protocol.ResponseError) *dshackle.NativeCallReplyItem
```
tonatoz: `ErrorItem` and `errorItemRenderer` encode the same per-kind vocabulary twice: for JSON-RPC and REST `ErrorItem` is `noUpstreamErrorItem` while the renderer is `nativeCallErrorItem`; for gRPC both wrap `grpcNativeCallErrorItem`. Keeping only the adapter method — `errorItem(requestID, err, errorAsIs)`, called by `sendReply` and by the pre-dispatch path (which additionally sets `UpstreamId = flow.NoUpstream`) — would leave one definition per kind.

internal/server/emerald/native_call_adapter_grpc.go
```diff
@@ -0,0 +1,93 @@
...
+
+	// arity is not on the gRPC wire - only the spec knows a method's shape.
+	// Unknown methods answer precisely here instead of "no available upstreams",
+	// and server streams belong to NativeSubscribe (the flow would route them
+	// as subscriptions).
+	// A method without the grpc connector is not a gRPC method on this chain:
+	// letting it through would route proto bytes to an HTTP connector.
+	specMethod := specs.GetSpecMethod(chain.MethodSpec, item.GetMethod())
+	if specMethod == nil || !slices.Contains(specMethod.GetApiConnectorTypes(), specs.GrpcConnector) {
+		return nil, a.ErrorItem(item.GetId(), protocol.NotSupportedMethodError(item.GetMethod()))
```
tonatoz: The `GrpcConnector` requirement is the right check, and it makes an existing asymmetry visible: the gRPC ingress (`chain_ingress.go:162-165`) accepts any method the spec knows, and `flow.getMethodConnector` (`internal/upstreams/flow/request_processor.go:175-179`) then picks the first declared connector the upstream has — so proto bytes for an HTTP-only method still reach the HTTP connector through that path. Pre-existing and out of scope here, but worth a follow-up issue.

tonatoz (changes requested):

internal/server/emerald/grpc_blockchain.go
KirillPamPam: Fixed

internal/server/emerald/grpc_blockchain.go
KirillPamPam: Added a comment

internal/server/emerald/native_call_adapter_grpc.go
KirillPamPam: Fixed

internal/server/emerald/native_call_adapter.go
KirillPamPam: grpc unary requests are not streamed, moreover we send trailers with the response item and callers should handle them in the same way as we do in chain ingress in nodecore. So everything is fine

internal/server/emerald/native_call_adapter.go
KirillPamPam: Fixed

internal/server/emerald/native_call_adapter.go
KirillPamPam: Fixed

internal/server/emerald/native_call_adapter_grpc.go
KirillPamPam: Added an additional check in chain ingress to handle a grpc connector

tonatoz (approved):

PR #357, opened 2026-09-02
Dedicated proof lower-bound pipeline for op-reth

tonatoz:
## Summary

Proofs get their own lower-bound detector. `EvmProofLowerBoundDetector` asks, in order and on every cycle: `debug_proofsSyncStatus` (op-reth historical proof store), then `eth_capabilities`, then the `eth_getProof` binary search. It also sets the label `historical_proofs=true`. One sync-status call replaces the binary search on an op-reth node.

The separate pipeline exists because `eth_capabilities` is unreliable for proofs. op-reth on op-sepolia reports `stateproofs.oldestBlock == head.number` while `--proofs-history` serves a window 129600 blocks deep; Ethereum geth reports a sane value. The proof detector therefore trusts `stateproofs.oldestBlock` only when it is below the `head.number` of the same report, and goes to the search otherwise. `ProofBound` is gone from the generic `EvmLowerBoundDetector` entirely.

The window's upper edge is not published. Measurements on live nodes (op-reth v2.4.2 on optimism and op-sepolia, reth v2.5.2 on base-sepolia) show `latest == head +- 2 blocks` with a sliding window of 129600 blocks, and `earliest == head - 129600`. The upper edge therefore follows from the head that nodecore already tracks, and a separate bound type would only duplicate it. This PR carried such a type in earlier revisions; it is gone from nodecore, and `drpcorg/public` removed `LOWER_BOUND_PROOF_UPPER` in v1.3.1 with enum 11 reserved and unused.

## Changes

- **deps**: `github.com/drpcorg/public` v1.2.1 -> v1.3.1, which ships the `debug_proofsSyncStatus` spec entry with `cacheable: false`.
- **evm bounds**: new `EvmProofLowerBoundDetector` in `proof_bound.go` with the three-stage pipeline and no cached verdicts. Each cycle is request -> parse -> result or error: a rejected or malformed sync status simply falls through to the next stage, and an upstream without the method pays one rejected request per 3 minutes. That replaces the mutex, the `unsupported` flag and the hourly re-probe timer of the deleted `EvmProofsSyncStatus` (review comment on `proofs_sync_status.go:37`).
- **evm bounds**: `ProofBound = earliest` (`earliest = 0` is coerced to 1, because 0 reads as "unknown" to routing). An empty window (`latest = 0`, or `earliest > latest` while the store initialises) is not a bound and falls through.
- **evm bounds**: `EvmLowerBoundDetector` loses the proof stage and the `proofsSyncStatus` field; its order is now `eth_capabilities` -> gold bound -> search. The shared JSON-RPC helper (`call`, `fetchLatestHeight`) moved into an embedded `evmRpcClient` value both detector types use.
- **evm bounds**: the `eth_capabilities` snapshot now carries the reported `head.number` (0 when absent or unparseable). A report without a head cannot be validated, so its `stateproofs` value is unusable for the proof bound.
- **evm bounds**: `parseEvmBlockNumber` accepts both shapes a block number arrives in - a quoted hex string (`"0x1a"`) and a bare decimal number (`26`). Live nodes answer decimal, which the hex parser would read as `0x26`.
- **wiring**: the proof detector is attached whenever the chain spec has `eth_getProof`. The `hasMethod("debug_proofsSyncStatus")` gate is dropped for it, because the detector asks the method every cycle and treats a rejection as "next stage". The `historical_proofs` label detector now uses the same `eth_getProof` gate: the label states what backs that method, and `debug_proofsSyncStatus` sits in the base `eth-json-rpc` spec every EVM chain inherits, so its gate excluded nothing while chains that disable `eth_getProof` (viction, hyperliquid) still ran the detector.
- **labels**: `EthHistoricalProofsLabelsDetector` publishes `historical_proofs=true` when the method answers with a window (even an empty one - the store exists), `false` when the upstream definitely lacks the method, and nothing on transient or malformed answers, so the last verdict stands. Unchanged by the refactor.
- **methods**: `debug_proofsSyncStatus` joins `probedMethods`, so a node that lists the `debug` module but is not op-reth stops advertising it.

No protocol, flow, chain-aggregation or emerald mapping changes: `ProofBound` behaves like every other lower bound.

## Verification

`go build ./...` is clean, `go vet ./internal/upstreams/...` is clean, `gofmt -l internal` is empty, and `go test ./internal/upstreams/lower_bounds/... ./internal/upstreams/chains_specific/... ./internal/upstreams/labels/...` passes.

Eight tests in `proof_bound_test.go` cover the pipeline: a sync-status window yields the bound with one upstream call; decimal input with `earliest = 0` yields 1; a rejected sync status (`-32601`, textual "Method not found", unclassified error) falls to `eth_capabilities` and is asked again on the next cycle (2 requests over 2 cycles, no memoized verdict); malformed and empty windows fall to capabilities; `stateproofs.oldestBlock == head.number` and a report without a head both fall to the search; `disabled: true` publishes nothing; a detector without capabilities searches directly.

Runtime check against a fake op-reth on chain optimism (head 5000000, window 129600):

- Sync-status mode: `upstream 'fake-opreth' lower bound of type PROOF is 4870400` - exactly `head - 129600`. `eth_getProof` call count on the fake node: 0.
- op-sepolia shape (sync status answered with -32601, `eth_capabilities` with `stateproofs.oldestBlock == head.number`): the log shows `upstream 'fake-opreth' eth_capabilities reports proofs from 5000000 with head 5000000, ignoring it for the proof bound`, then `lower bound of type PROOF is 4000000` - the height the `eth_getProof` search finds on the fake node, never the head.
- After two 3-minute cycles against that node, `debug_proofsSyncStatus` was requested twice, so nothing is memoized in the production wiring.
- `upstream 'fake-opreth' label of historical_proofs is true` in sync-status mode.
- `debug_proofsSyncStatus` through the HTTP entrypoint returned `{"earliest": 4870400, "latest": 5000000}`, and each call raised the upstream's call counter, so the response is not cached.

## Notes for reviewers

- The implementation plan is deleted and replaced by `docs/superpowers/specs/2026-09-07-proof-lower-bound-design.md`, which describes the shipped three-stage detector. The plan specified `LOWER_BOUND_PROOF_UPPER`, a memoizing sync-status source and a gRPC mapping, none of which shipped.
- The plan asked for a mirrored proto edit in `emerald-grpc/proto/blockchain.proto`. That submodule was removed in #353, nothing imports `github.com/drpcorg/emerald-grpc`, and the authoritative proto now comes from `drpcorg/public`, so the mirror was skipped.
- Nothing downstream needs a new enum: `LOWER_BOUND_PROOF` already exists in every client.


internal/upstreams/lower_bounds/evm_bounds/proofs_sync_status.go
```diff
@@ -0,0 +1,179 @@
...
+// unparseable body, is remembered as unsupported and re-asked hourly; transient failures
+// leave the verdict alone.
+type EvmProofsSyncStatus struct {
+	upstreamId      string
+	chain           *chains.ConfiguredChain
+	internalTimeout time.Duration
+	connector       connectors.ApiConnector
+	reprobeInterval time.Duration
+
+	mu            sync.Mutex
```
KirillPamPam: Let's omit these fields. It could be come down to the simple logic: request -> request -> parsing -> error/result

tonatoz: Done: the memoizing fields are gone. `proofs_sync_status.go` is deleted and the proof stage now lives in `EvmProofLowerBoundDetector.detectFromProofsSyncStatus` as plain request -> parse -> result/error, with no mutex, no `unsupported` verdict and no re-probe timer. An upstream without the method rejects one call per 3-minute cycle and the detector moves to the next stage.

While removing the cache I also split proofs off the generic detector: `eth_capabilities` is wrong for them on op-reth (op-sepolia reports `stateproofs.oldestBlock == head.number` while serving a 129600-block window), so the new pipeline is `debug_proofsSyncStatus` -> `eth_capabilities` trusted only below the reported `head.number` -> `eth_getProof` search. Updated description and tests in 148c2ce.

KirillPamPam (approved):

PR #365, opened 2026-09-10
Syncing guard

KirillPamPam:
# Pause EVM heads while the node is syncing

## Summary

A syncing EVM node does two things that hurt us. Newer Arbitrum Nitro releases answer
`eth_syncing` with an object we did not recognise, so the health probe reported the node as
`Available` while it was replaying history. And a syncing node with a websocket head streams
`newHeads` at replay speed: fork choice already drops the heads of a non-available upstream,
so every one of those events was parsed and discarded for nothing.

This PR fixes the probe, stops the head processor of an EVM upstream for as long as the
`eth_syncing` probe says the node is syncing, and teaches the websocket head-liveness
detector that a paused head is not a stalled one. dshackle does the same thing on its head
level (`GenericWsHead.onSyncingNode`); here the guard lives in the upstream state pipeline,
which already separates the probe verdict from the lag-derived one.

### `eth_syncing` probe recognises the new Nitro shape

`EthSyncingValidator` returns `Syncing` when the result object carries `executionSyncTarget`
and `messageOfLastBlock`, the fields newer Nitro emits while the execution layer is behind
its target:

```json
{
  "blockNum": 43327714,
  "consensusMaxMessageCount": 92338320,
  "executionSyncTarget": 92338319,
  "messageOfLastBlock": 43327714
}
```

The existing fallback for unknown structured payloads (`Available`) is unchanged.

### Head pause driven by the probe, not by lag

`GenericUpstream.processStateEvents` already keeps `baseAvail`, the availability reported by
health probes, separately from the effective status that the chain supervisor's lag observer
can downgrade to `Syncing`. Only probe events (those without a `Lag`) update `baseAvail`, so
that is where the guard sits:

- when `baseAvail` transitions into `Syncing`, the head event processor is stopped, which
  unsubscribes the head and stops its no-updates timer;
- when it transitions to anything else (`Available` or `Immature`), the head event processor
  is started again, which refetches the latest block and resubscribes.

A node that merely fell behind by lag keeps its head. The fatal-error branch resets the pause
bookkeeping because `PartialStop` and `Resume` restart the head outside the loop.

The guard is opt-in per chain family through a new `ChainSpecific.PauseHeadWhileSyncing()`.
The EVM specific returns true; every other specific, Tron included, returns false. The upstream
captures the answer as a bool at construction, so it does not need the chain-specific itself.

`GenericHeadProcessor.UpdateHead` (manual heads from the integrity processor) is now a
non-blocking send: with a paused head nobody drains that channel, and the previous blocking
send could hang a request once the buffer filled.

### Head-liveness detector: a gap in observation is not a stall

`WsHeadLivenessCapDetector` gates `WsCap` on the head advancing. Its tracker learns the block
time as the largest interval seen (it never shrinks) and doubles it on every stall timeout. A
paused head, or a dropped socket, used to poison that estimate twice: the timer kept firing
during the pause (backoff up to the 10-minute cap, a 60-minute stall window), and the first
head after resume was measured against the pre-pause timestamp. After one long sync the
detector was effectively blind for the rest of the process lifetime.

Now:

- `blocks.HeadEvent` is a sealed interface. The former struct is `HeadBlockEvent`; the new
  `HeadStateEvent{Running bool}` is published by the head processor on `Start` and `Stop`.
  One ordered stream keeps blocks and lifecycle changes in the order they happened, so the
  detector can tell a block that arrived before a stop from one that arrived after a start.
  The head event processor forwards only blocks to the state pipeline.
- The tracker gets `suspend()`: liveness is retracted, the consecutive run and the last
  height/timestamp are forgotten, `measuredBlockTime` and the forward-gap cooldown are left
  alone. The detector calls it on websocket disconnect and on head stop.
- The stall timer is armed only while observing (socket connected and head running) and held
  stopped otherwise. On resume it is re-armed with the unchanged window.
- The first head after a suspend is a new baseline, so the height jump accumulated while not
  looking is not mistaken for a forward gap and no cooldown is imposed. A fresh run of three
  heads brings the cap back.

**Behaviour change:** after a websocket reconnect the cap no longer returns on its own; the
head must produce a fresh consecutive run first. The test that asserted the old behaviour now
asserts this one.

## Tests

- `eth_health_validator_test.go`: the Nitro payload above yields `Syncing`.
- `upstream_test.go`: probe `Syncing` pauses the head, a repeated verdict is a no-op,
  `Available` and `Immature` both resume; lag-driven `Syncing` does not pause; without the flag
  nothing is paused; the pause bookkeeping resets across a fatal error.
- `head_processor_internal_test.go`: `Start`, a block and `Stop` publish in order on one
  subscription; `UpdateHead` does not block on a full channel with no reader.
- `block_event_processor_test.go`: state events are not forwarded to the state pipeline.
- `ws_head_liveness_detector_internal_test.go`: `suspend` keeps the block time and requires a
  fresh run without a cooldown; the first observation after a suspend does not measure the gap;
  a height jump after a suspend is a baseline, not a forward gap.
- `ws_head_liveness_detector_test.go`: a stopped head retracts the cap, no timeout fires during
  the pause, and after restart the cap returns on a fresh run at a much higher height while a
  real stall is still caught with the original window; the same for a websocket disconnect;
  reconnect alone does not restore the cap.

### DIFF
--- a/internal/upstreams/blocks/head_processor.go
+++ b/internal/upstreams/blocks/head_processor.go
@@ -59,6 +59,9 @@ type GenericHeadProcessor struct {
 	headNoUpdatesTimeout time.Duration
 	subManager           *utils.SubscriptionManager[HeadEvent]
 	manualHeadChan       chan protocol.Block
+	// drainDone is closed when the current run's drain goroutine exits. Stop waits on it so
+	// that HeadStateEvent{Running: false} is published after the last block of the run.
+	drainDone chan struct{}
 }
 
 func NewGenericHeadProcessor(
@@ -116,7 +119,10 @@ func (h *GenericHeadProcessor) Start() {
 		h.lastUpdate.Store(time.Now())
 		h.subManager.Publish(HeadStateEvent{Running: true})
 
+		drainDone := make(chan struct{})
+		h.drainDone = drainDone
 		go func() {
+			defer close(drainDone)
 			timeout := time.NewTimer(h.headNoUpdatesTimeout)
 			for {
 				select {
@@ -147,8 +153,14 @@ func (h *GenericHeadProcessor) Start() {
 	})
 }
 
+// Stop cancels the drain goroutine and waits for it before publishing the stop, so nothing
+// of this run is published after HeadStateEvent{Running: false}: a late subscriber's replay
+// then reports a paused head exactly when the head is paused.
 func (h *GenericHeadProcessor) Stop() {
 	h.lifecycle.Stop()
+	if h.drainDone != nil {
+		<-h.drainDone
+	}
 	h.head.Stop()
 	h.subManager.Publish(HeadStateEvent{Running: false})
 }
--- a/internal/upstreams/blocks/head_processor.go
+++ b/internal/upstreams/blocks/head_processor.go
@@ -24,9 +24,27 @@ type HeadProcessor interface {
 	Subscribe(name string) *utils.Subscription[HeadEvent]
 }
 
-type HeadEvent struct {
+// HeadEvent is what a head processor publishes to its subscribers: head blocks interleaved
+// with lifecycle changes, in the order they happened. One stream keeps them ordered, so a
+// consumer can tell a block that arrived before a stop from one that arrived after a start.
+type HeadEvent interface {
+	headEvent()
+}
+
+// HeadBlockEvent carries a new head of the upstream.
+type HeadBlockEvent struct {
 	HeadData protocol.Block
 }
+
+// HeadStateEvent reports that the head processor started or stopped observing the node.
+// Silence after Running == false is not a stall: nothing is being observed.
+type HeadStateEvent struct {
+	Running bool
+}
+
+func (HeadBlockEvent) headEvent() {}
+
+func (HeadStateEvent) headEvent() {}
 
 type GenericHeadProcessor struct {
 	upstreamId           string
@@ -87,6 +105,7 @@ func (h *GenericHeadProcessor) Start() {
 	h.lifecycle.Start(func(ctx context.Context) error {
 		h.head.Start()
 		h.lastUpdate.Store(time.Now())
+		h.subManager.Publish(HeadStateEvent{Running: true})
 
 		go func() {
 			timeout := time.NewTimer(h.headNoUpdatesTimeout)
@@ -102,14 +121,14 @@ func (h *GenericHeadProcessor) Start() {
 					if ok {
 						log.Debug().Msgf("got a new head of upstream %s - %d", h.upstreamId, block.Height)
 						h.lastUpdate.Store(time.Now())
-						h.subManager.Publish(HeadEvent{HeadData: block})
+						h.subManager.Publish(HeadBlockEvent{HeadData: block})
 					}
 				case manualBlock := <-h.manualHeadChan:
 					if manualBlock.Height > h.head.GetCurrentBlock().Height {
 						log.Debug().Msgf("got a new manual head of upstream %s - %d", h.upstreamId, manualBlock.Height)
 						h.lastUpdate.Store(time.Now())
 						h.head.UpdateHead(manualBlock)
-						h.subManager.Publish(HeadEvent{HeadData: manualBlock})
+						h.subManager.Publish(HeadBlockEvent{HeadData: manualBlock})
 					}
 				}
 				timeout.Reset(h.headNoUpdatesTimeout)
@@ -122,10 +141,16 @@ func (h *GenericHeadProcessor) Start() {
 func (h *GenericHeadProcessor) Stop() {
 	h.lifecycle.Stop()
 	h.head.Stop()
+	h.subManager.Publish(HeadStateEvent{Running: false})
 }
 
 func (h *GenericHeadProcessor) UpdateHead(height, slot uint64) {
-	h.manualHeadChan <- protocol.NewBlockWithHeights(height, slot)
+	select {
+	case h.manualHeadChan <- protocol.NewBlockWithHeights(height, slot):
+	default:
+		// nobody drains the channel while the head is paused; a dropped manual head
+		// costs nothing, the next real head supersedes it
+	}
 }
 
 func createHead(
--- a/internal/upstreams/blocks/head_processor.go
+++ b/internal/upstreams/blocks/head_processor.go
@@ -22,6 +22,11 @@ type HeadProcessor interface {
 	UpdateHead(height, slot uint64)
 
 	Subscribe(name string) *utils.Subscription[HeadEvent]
+	// SubscribeWithReplay also delivers the last published event first. On this stream that
+	// event is HeadStateEvent{Running: false} exactly when the head is stopped - Stop
+	// publishes it last and nothing else is published until Start - so a late subscriber
+	// learns whether the head is currently paused.
+	SubscribeWithReplay(name string) *utils.Subscription[HeadEvent]
 }
 
 // HeadEvent is what a head processor publishes to its subscribers: head blocks interleaved
@@ -95,6 +100,10 @@ func (h *GenericHeadProcessor) GetCurrentBlock() protocol.Block {
 
 func (h *GenericHeadProcessor) Subscribe(name string) *utils.Subscription[HeadEvent] {
 	return h.subManager.Subscribe(name)
+}
+
+func (h *GenericHeadProcessor) SubscribeWithReplay(name string) *utils.Subscription[HeadEvent] {
+	return h.subManager.SubscribeWithReplay(name)
 }
 
 func (h *GenericHeadProcessor) Running() bool {
--- a/internal/upstreams/blocks/head_processor.go
+++ b/internal/upstreams/blocks/head_processor.go
@@ -156,12 +156,16 @@ func (h *GenericHeadProcessor) Start() {
 // Stop cancels the drain goroutine and waits for it before publishing the stop, so nothing
 // of this run is published after HeadStateEvent{Running: false}: a late subscriber's replay
 // then reports a paused head exactly when the head is paused.
+//
+// The head is stopped before the wait: the drain goroutine may be parked inside
+// OnNoHeadUpdates, in a resubscribe whose request honours the head's lifecycle context, and
+// head.Stop() cancelling that context is what lets the goroutine return and exit.
 func (h *GenericHeadProcessor) Stop() {
 	h.lifecycle.Stop()
+	h.head.Stop()
 	if h.drainDone != nil {
 		<-h.drainDone
 	}
-	h.head.Stop()
 	h.subManager.Publish(HeadStateEvent{Running: false})
 }
 
--- a/internal/upstreams/caps/ws_head_liveness_detector.go
+++ b/internal/upstreams/caps/ws_head_liveness_detector.go
@@ -121,6 +121,19 @@ func (h *headLivenessTracker) onTimeout() bool {
 	return h.live
 }
 
+// suspend records that observation stopped: the head is paused or the socket is down.
+// Liveness is retracted and the next observe becomes a new baseline that starts a fresh run,
+// so the height jump accumulated while not looking is not mistaken for a forward gap and no
+// cooldown is imposed. measuredBlockTime is untouched: nothing was learned about the chain,
+// and the gap is ours, not the node's. Returns the (now false) verdict.
+func (h *headLivenessTracker) suspend() bool {
+	h.live = false
+	h.count = 0
+	h.haveLast = false
+	h.haveLastBlockTime = false
+	return h.live
+}
+
 // timeout is the stall window: how long the detector waits for head progress before calling onTimeout.
 func (h *headLivenessTracker) timeout() time.Duration {
 	return h.measuredBlockTime * checkedBlocksUntilLive * timeoutMultiplier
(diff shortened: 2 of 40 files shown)


internal/upstreams/validations/eth_validations/eth_health_validator.go
```diff
@@ -123,6 +123,11 @@ func (e *EthSyncingValidator) Validate() protocol.AvailabilityStatus {
 		root.Get("syncTargetMsgCount").Exists() {
 		return protocol.Syncing
 	}
+	// newer Arbitrum Nitro releases report sync progress with these fields instead
+	if root.Get("executionSyncTarget").Exists() &&
```
tonatoz, line 127: This adds a third key-existence branch, and the fallback at line 131 still returns `Available` for any unrecognized non-empty object. Arbitrum documents that the `eth_syncing` output of an out-of-sync Nitro node is not a stable API and can change between versions (https://docs.arbitrum.io/build-decentralized-apps/arbitrum-vs-ethereum/rpc-methods), so the next field rename brings the bug back silently.

Consider inverting the default after the geth `currentBlock`/`highestBlock` branch: treat any non-empty object as `Syncing` and drop both Nitro key checks. A synced client answers `false` per the spec, so a false positive keeps an upstream out of rotation instead of routing traffic to a node that is replaying history.

internal/upstreams/blocks/head_processor.go
```diff
@@ -119,13 +153,29 @@ func (h *GenericHeadProcessor) Start() {
...
+// then reports a paused head exactly when the head is paused.
+//
+// The head is stopped before the wait: the drain goroutine may be parked inside
+// OnNoHeadUpdates, in a resubscribe whose request honours the head's lifecycle context, and
+// head.Stop() cancelling that context is what lets the goroutine return and exit.
 func (h *GenericHeadProcessor) Stop() {
 	h.lifecycle.Stop()
 	h.head.Stop()
+	if h.drainDone != nil {
+		<-h.drainDone
```
tonatoz: The doc comment assumes `head.Stop()` releases a drain goroutine parked in `OnNoHeadUpdates`, which `resubscribingHead` in the new test also models. The real `SubscriptionHead.OnNoHeadUpdates` (head.go:231-235) is `w.Stop(); w.Start()`, so that does not hold.

If this `Stop` interleaves between those two calls: `h.head.Stop()` finds the head lifecycle already stopped (no-op CAS, `Unsubscribe` on an already-released opId), then `w.Start()` passes its CAS and opens a fresh subscription on a new context derived from the upstream parent context, which this `Stop` has no handle on. The drain goroutine then exits on its own context, `drainDone` closes, `HeadStateEvent{Running: false}` is published, and the `newHeads` subscription to the syncing node stays open with its reader parked on `w.headsChan <- block`.

Two consequences: the pause does not actually stop the work it is meant to stop, and on resume `h.head.Start()` logs "already running", skips `getLatestBlock`, and the first block handed to the new drain is the one captured before the pause, which is what the new `select` in head.go is meant to prevent.

Also note that `<-h.drainDone` runs on the `processStateEvents` goroutine and holds the aggregator mutex. The wait is bounded by `wsWriteTimeout = 30s` (ws_processor.go:33, `sendWsRequest` does not await the node's reply), so `PartialStop`, `Resume`, `Stop` and state-event processing for this upstream can stall for that long, after which `stateChan` (1000) fills and emitters block.

Suggested: make the head's stop final so `OnNoHeadUpdates` cannot resubscribe after a `Stop` (a mutex plus a `stopped` flag, or drive the resubscribe from the processor instead of from inside the head). Calling `h.head.Stop()` once more after `<-h.drainDone` closes the resurrection window, and bounding the wait with a `select` keeps the state loop independent of a network call.

internal/upstreams/blocks/head_processor.go
```diff
@@ -36,6 +59,9 @@ type GenericHeadProcessor struct {
 	headNoUpdatesTimeout time.Duration
 	subManager           *utils.SubscriptionManager[HeadEvent]
 	manualHeadChan       chan protocol.Block
+	// drainDone is closed when the current run's drain goroutine exits. Stop waits on it so
+	// that HeadStateEvent{Running: false} is published after the last block of the run.
+	drainDone chan struct{}
```
tonatoz: `drainDone` is written in the body of `lifecycle.Start` and read in `Stop` without synchronization. It is safe today only because every caller goes through the new mutex in `UpstreamProcessorAggregator`, an invariant that lives in another package and is not stated here. The race detector will not catch it: no test drives `Start`/`Stop` from two goroutines. `utils.Atomic[chan struct{}]` (already used in this struct for `lastUpdate`) would remove the cross-package coupling.

internal/upstreams/blocks/head_processor.go
```diff
@@ -119,13 +153,29 @@ func (h *GenericHeadProcessor) Start() {
...
+	}
+	h.subManager.Publish(HeadStateEvent{Running: false})
 }
 
 func (h *GenericHeadProcessor) UpdateHead(height, slot uint64) {
-	h.manualHeadChan <- protocol.NewBlockWithHeights(height, slot)
+	select {
+	case h.manualHeadChan <- protocol.NewBlockWithHeights(height, slot):
+	default:
+		// nobody drains the channel while the head is paused; a dropped manual head
```
tonatoz, line 177: The non-blocking send is right, but the drop is silent even when the processor is running and the 100-slot buffer is merely full, which makes a lost manual head undiagnosable. A `log.Debug` or a counter here would keep it visible.

internal/upstreams/caps/ws_head_liveness_detector.go
```diff
@@ -170,7 +190,7 @@ func (d *WsHeadLivenessCapDetector) DetectCaps(ctx context.Context) <-chan mapse
 		close(out)
```
tonatoz, line 190: Two notes on this subscription.

1. The name is fixed, while `DetectCaps` is re-entered whenever the cap processor restarts (`PartialStop` then `Resume`, handled on one supervisor goroutine). `SubscribeWithReplaySize` calls `log.Panic` on a name collision, and `Stop` only cancels the context, so the previous run's deferred `Unsubscribe` may not have run yet. `HeadEventProcessor.Start` works around the same hazard with a uuid; the same window exists here and in `cap_event_processor.go:36`, `methods_event_processor.go:24`, `lower_bound_processor.go:40`. Pre-existing, but worth fixing at the source (synchronous `Unsubscribe`, or an idempotent `Subscribe`) rather than spreading uuids.

2. The behaviour change deserves an explicit sign-off: after every ws reconnect or head pause, `WsCap` now needs a fresh run of `checkedBlocksUntilLive = 3` heads. That is roughly 24-36 s on a 12 s chain and minutes on slow chains, repeated per reconnect, where previously the cap returned immediately.

internal/upstreams/caps/ws_head_liveness_detector.go
```diff
@@ -183,6 +203,10 @@ func (d *WsHeadLivenessCapDetector) DetectCaps(ctx context.Context) <-chan mapse
 			measuredBlockTime: d.expectedBlockTime,
 		}
 		var wsConnected, headLive bool
+		// A head that is already paused announces itself through the replayed
+		// HeadStateEvent{Running: false}; anything else on the stream means it is running.
```
tonatoz: The optimistic default holds only while a replayed event exists. If the detector subscribes before the head processor has ever published and the socket connects first, `observing()` is true against a head that is not running, the stall timer fires, and `onTimeout` backs off `measuredBlockTime`, which is the poisoning this PR removes. The window is small because `Resume()` starts the head before the cap processor (upstream.go:274-281). Taking the initial value from `headProcessor.Running()`, or arming the timer only after the first `HeadStateEvent`, removes the assumption.

internal/upstreams/event_processors/block_event_processor.go
```diff
@@ -95,7 +96,9 @@ func (h *HeadEventProcessor) Start() {
 	h.lifecycle.Start(func(ctx context.Context) error {
 		h.headProcessor.Start()
 
-		headSub := h.headProcessor.Subscribe(fmt.Sprintf("%s_head_updates", h.upstreamId))
+		// unique per start: the previous run releases its subscription asynchronously, and
+		// a pause followed by an immediate resume must not clash with it
+		headSub := h.headProcessor.Subscribe(fmt.Sprintf("%s_head_updates_%s", h.upstreamId, uuid.NewString()))
```
tonatoz, line 101: The uuid fixes a real collision window, but the underlying cause is that `Stop` only cancels the context while `Unsubscribe` runs later in the goroutine, and `SubscribeWithReplaySize` panics via `log.Panic` on a duplicate name. The same window exists for the fixed names in `cap_event_processor.go:36`, `methods_event_processor.go:24`, `lower_bound_processor.go:40` and `ws_head_liveness_detector.go:190`. Fixing it in the subscription manager or in the stop path would cover all of them.

internal/upstreams/upstream_events.go
```diff
@@ -96,6 +102,7 @@ func (u *GenericUpstream) processStateEvents(ctx context.Context, initialValid b
 				}
 				if stateEvent.Lag == nil {
 					baseAvail = stateEvent.Status
+					headPaused = u.toggleHeadOnSyncing(baseAvail, headPaused)
```
tonatoz: `headPaused` duplicates state that another goroutine also owns: the supervisor calls `PartialStop`/`Resume` on the same processor (upstream_supervisor.go:183-189). No divergence occurs today only because `Resume()` is always preceded by a `FatalErrorUpstreamStateEvent` that resets the flag. If a `Resume()` ever lands while the flag is true, the head runs while the probe reports `Syncing` until the next non-`Syncing` verdict, with no log to show it.

`StartProcessor`/`StopProcessor` are already idempotent through the CAS in `GenericLifecycle`, so the flag only suppresses a log line. Reading the processor's `Running()` instead would make the desync impossible.

internal/upstreams/upstream_events.go
```diff
@@ -47,6 +50,9 @@ func (u *GenericUpstream) processStateEvents(ctx context.Context, initialValid b
 				log.Warn().Msgf("upstream '%s' settings are invalid, it will be stopped", u.id)
 				eventType = &protocol.RemoveUpstreamEvent{}
 				validUpstream = false
+				// PartialStop takes the head down with everything else and Resume brings it
+				// back outside this loop, so the pause bookkeeping starts over
+				headPaused = false
```
tonatoz: The reset is correct but carries an implicit invariant: the branch that precedes `Resume()` is `ValidUpstreamStateEvent` (line 57), and it does not touch `headPaused`. This works because `validUpstream` can only go back to true after this branch has already reset the flag. Worth stating in the comment, or resetting in both places, so a future reordering does not break the guard silently.

internal/upstreams/blocks/head_processor.go
```diff
@@ -119,13 +153,29 @@ func (h *GenericHeadProcessor) Start() {
...
+// The head is stopped before the wait: the drain goroutine may be parked inside
+// OnNoHeadUpdates, in a resubscribe whose request honours the head's lifecycle context, and
+// head.Stop() cancelling that context is what lets the goroutine return and exit.
 func (h *GenericHeadProcessor) Stop() {
 	h.lifecycle.Stop()
 	h.head.Stop()
+	if h.drainDone != nil {
+		<-h.drainDone
+	}
+	h.subManager.Publish(HeadStateEvent{Running: false})
```
tonatoz, line 170: `HeadStateEvent{Running: false}` is published unconditionally, including when the lifecycle was already stopped or never started, so the stream can carry a `false` with no preceding `true`. Harmless for the current consumer because `suspend()` is idempotent, but gating the publish on the `lifecycle.Stop()` CAS result would keep the stream an accurate transition log.

Separately: while the head is paused, `headsMetric` freezes, so a paused head looks identical to a stuck head on dashboards. Worth accounting for in alerting.

internal/upstreams/blocks/head_processor.go
```diff
@@ -119,13 +153,29 @@ func (h *GenericHeadProcessor) Start() {
 	})
 }
```
tonatoz, line 157: Test note: `TestGenericHeadProcessorPublishesStateAroundBlocks`, `TestGenericHeadProcessorStopPublishesStateAfterTheLastBlock` and `TestGenericHeadProcessorStopReleasesDrainParkedInResubscribe` build `GenericHeadProcessor` with struct literals and a head stub whose `Stop()` releases the parked resubscribe. That is why the divergence with the real `SubscriptionHead` described above is invisible to the suite. Also `resubscribingHead.Stop()` closes a channel, so a second `Stop` panics, which couples the stub to being called exactly once.

internal/upstreams/validations/eth_validations/eth_health_validator.go
KirillPamPam: I suggest to leave it as it. We can't be sure what an object might mean and treat it as `Syncing` by default is not a good idea

internal/upstreams/blocks/head_processor.go
KirillPamPam: Agree, fixed

internal/upstreams/caps/ws_head_liveness_detector.go
KirillPamPam: 1. Will fix it separately in the whole app
2. It's ok

internal/upstreams/event_processors/block_event_processor.go
KirillPamPam: Will be done separately

internal/upstreams/blocks/head_processor.go
KirillPamPam: It's harmless, suggest to leave it

tonatoz (approved):

PR #371, opened 2026-09-15
Cosmos + EVM specific

KirillPamPam:
# EVM connectors on Cosmos chains (Injective)

## Summary

Cosmos SDK chains with an EVM module serve the Ethereum JSON-RPC (and its websocket) next to the
CometBFT RPC, the LCD and the SDK gRPC API. Injective is the first one we route. Until now a cosmos
upstream could only carry `tendermint`, `rest` or `grpc` connectors; a `json-rpc` one was refused
by the cosmos factory, and even if it had been accepted, the EVM chain validator would have compared
`eth_chainId` against the cosmos network name (`injective-888`) and refused the upstream.

This PR lets an Injective upstream mix `json-rpc` / `websocket` with the cosmos connectors, gives
the EVM side its own chain ids, and separates the specific that runs the probes from the specific
that drives the head - on a mixed-API upstream they can belong to different API families.

Requires `github.com/drpcorg/public` **v1.4.0** (bumped here), which adds the `cosmos-evm` bundle
(`eth` + `cosmos`), points Injective at it, and declares the EVM ids in `chains.yaml`:

```yaml
- id: Testnet
  chain-id: injective-888
  chain-ids:
    cosmos: injective-888
    eth: 0x59f
  net-version: 888
```

## What was verified against a live node

Injective testnet, the same block fetched from all four APIs:

| API | height | hash | parent |
|---|---|---|---|
| CometBFT `/block` | 139732091 | `57D9…E12F` (hex) | `E50F…BC16` |
| LCD `blocks/{h}` | 139732091 | same, base64 | same |
| gRPC `GetBlockByHeight` | 139732091 | same, raw bytes | same |
| EVM `eth_getBlockByNumber` | 139732091 | same, `0x` hex | same |

The EVM `hash` is the CometBFT block-id hash and `parentHash` is `last_block_id.hash`, so every
flavor reduces to the same `HashId` and cross-upstream lag detection works unchanged. `latest`,
`finalized` and `safe` answer the same height in one batch request. The chain ids differ per API:
`eth_chainId` is `0x59f` (mainnet `0x6f0`), `net_version` is `888` (mainnet `1776`) - and on
testnet `net_version` is *not* the decimal of `eth_chainId`, hence the explicit `net-version`.

### Per-family chain ids (`pkg/chains`)

- `ConfiguredChain.ChainIds` carries the new `chain-ids` map, keyed by `BlockchainType`.
- `ChainIdFor(t)` returns the family's id when declared, otherwise the chain's own `chain-id`.
  Every chain validator and every "is chain validation configured" check now reads through it -
  the cosmos/tendermint ones with `chains.Cosmos`, the EVM validator and the locally served
  `eth_chainId` with `chains.Ethereum`, and so on. No direct `.ChainId` reads remain outside the
  registry.
- `net_version` stays a single field: the loader derives it from the EVM chain id (`chain-ids.eth`
  when declared, else `chain-id`) unless `net-version` is explicit, the same rule as before.
- The loader rejects a `chain-ids` key that is not a blockchain type, and an entry for the chain's
  own type that contradicts `chain-id`.
- `GetChainByChainIdAndVersion` also matches chains that declare an id for the queried family, so
  the EVM validator's "actually it's X" log names Injective correctly.

### Cosmos factory (`cosmos_specific`)

`NewCosmosSpecific` maps `json-rpc` and `websocket` connectors straight to
`evm_specific.NewEvmChainSpecific`, the same way Tron's json-rpc connector is handled. It now takes
the full connector list and the manual labels, which the upstream factory passes through.

No override of the EVM logic was needed: finalized equals latest on Injective, the pruned-state
error (`height N is not available, lowest height is M`) is already in the EVM no-data list,
`eth_syncing`, `eth_getLogs`, `eth_getProof`, `eth_gasPrice` and `txpool_status` all answer as
on an Ethereum node.

Which connector drives the probes follows the existing rule: the internal-request connector is the
lowest type by the default-mode order, so with `json-rpc` present it is `json-rpc`, and the EVM
logic runs health, chain validation, labels and lower bounds. That is the right side to probe: the
EVM state is pruned independently of the CometBFT block store (the test node kept blocks from
height 0 but state only from 64899000), and only the EVM detectors see that.

### One specific per role (`upstream_factory`)

Until now the factory built one `ChainSpecific` from the internal-request connector and handed it
to every processor, the head processor included. The head connector is chosen by a different rule
(lowest type in `default` mode, highest in `strict`), so on a mixed-API upstream the two could
disagree: a strict-mode Injective upstream with `json-rpc` + `tendermint` + `rest` + `grpc` got the
EVM specific for probes and a `grpc` head connector, and the head processor tried to open
`eth_subscribe` as a gRPC server stream - the node rejected it and the upstream stayed headless.

The factory now builds two specifics. The **probe** specific comes from the internal-request
connector and runs every poll: validators, labels, lower bounds, method detection, cap detectors
and the finalized/safe block processor. That is where they all ran before, so nothing moves. The
**head** specific comes from the head connector and drives only the head processor (and
`PauseHeadWhileSyncing`), the one place the connector's protocol matters because a subscription is
opened on it. On the usual upstream both connectors are the same object and so is the specific. On
a mixed-API upstream each side speaks its own connector's protocol by construction: a `grpc` head
is the Cosmos gRPC head (polled, since the Cosmos gRPC API has no head stream), a `websocket` head
is the EVM `newHeads` subscription, `tendermint` is the CometBFT poll. The head-connector defaults
are unchanged.

### EVM peers validator

Injective returns `net_peerCount` as a JSON number (`41`), where an Ethereum node returns a hex
string. The validator now accepts both; before, `validate-peers` (the strict-mode default) marked
such an upstream unavailable.

### Docs

`05-upstream-config.md`: the `json-rpc` / `websocket` connector bullets describe the cosmos-evm
case and which connector runs the probes; `head-connector` states that the head follows its own
connector's API family; the chain-id validator row points at `chain-ids` / `net-version`; the
peer-count row notes the numeric form.

## Tests

- `pkg/chains/chain_ids_test.go`: lookup, fallback, lowercasing, loader errors, EVM-side lookup by
  id and version, and the testnet net-version case.
- EVM chain validator: Injective testnet validates against `0x59f` / `888`.
- Local request processor: `eth_chainId` / `net_version` on a cosmos-evm chain answer the EVM ids.
- Cosmos factory: `json-rpc` and `websocket` yield the EVM specific; the unsupported-connector test
  moved to `rest-indexer` / `rest-additional`.
- Upstream factory: an Injective upstream with a `json-rpc` internal connector and a `grpc` head gets
  the EVM probe specific and the Cosmos gRPC head specific; a single connector yields one shared
  object.
- Peers validator: numeric `net_peerCount`.
- The cosmos spec-name test now expects `cosmos-evm` for Injective.


EugeneDrpc (approved):
LGTM
`
В strict mode на обычном EVM-апстриме с json-rpc + websocket head-коннектор = websocket, а internal = json-rpc, так что после этого PR они уже не «один объект»: фабрика создаёт второй EvmChainSpecificObject, привязанный к ws. В итоге SubscriptionHead.getLatestBlock (вызывается после каждого (ре)сабскрайба) теперь шлёт eth_getBlockByNumber через websocket вместо HTTP, а ws-запросы не ретраятся. Это самая частая продовая конфигурация, поэтому стоит явно отразить изменение в описании PR и добавить тест: probe = json-rpc, head = websocket, оба specific-а типа EvmChainSpecificObject, но разные объекты.`

